PR #4, opened 2012-04-17
New COB interactive teleop package

but-spanel:


floweisshardt:
Thanks for the pull request, please just update some small issues:
-Rename the include/but_cob_teleop include and src/but_cob_teleop source directory to fit the new package name.
-The namespace inside the source files should also be changed from but_cob_teleop to cob_interactive_teleop.
-In the manifest there's still something mentioned about "RViz plugin for COB arm navigation".

but-spanel:
Done, the package should be ready...

PR #10, opened 2012-05-24
fuerte support, compatible with electric

abubeck:


floweisshardt:
i think the code is not doing what you want:

point_msg.time_from_start=rospy.Duration(manipulation_duration*point_nr)

assigns the manipulation duration for each point. I think it is better to specify the manipulation_duration as the overall time, the trajectory should be executed. An example

trajectory A has 1 point, trajectory B has 3 points, trajectory C has 100 points

Total execution time of the trajectories will be 
current version (fixed to 3sec per point): A=3sec, B=9sec, C=300sec
your version (manipulation_duration set per point, e.g. set to 3sec): A=3sec, B=9sec, C=300sec
suggested version (manipulation_duration set for whole trajectory, e.g. 3sec): A=3sec, B=3sec, C=3sec

PR #13, opened 2012-06-15
COB stops faster when you release the marker...

but-spanel:


floweisshardt:
We found the real bug in the base_controller causing the delay, so no need for fixing this in the interactive_teleop node.

PR #15, opened 2012-10-02
A new launch file that remaps output topic to safe_commands added

but-spanel:


cob_interactive_teleop/launch/interactive_teleop_safe.launch
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<launch>
+
+  <rosparam ns="cob_interactive_teleop" file="$(find cob_interactive_teleop)/config/interactive_teleop.yaml" command="load" />
+
+  <node name="cob_interactive_teleop" pkg="cob_interactive_teleop" type="interactive_teleop" output="screen">
+    <remap from="/base_controller/command" to="/base_controller/command_safe" />
```
floweisshardt: to prevent many launch files for just small configuration issues, there are launch file arguments (see http://www.ros.org/wiki/roslaunch/XML/arg) which can be used to configure included launch files from parent launch files or the console.

suggestion: introduce arg "base_topic" which is by default pointing to "/base_controller/command_safe" but can be overwritten by setting the argument outside of the launch file.

Using this approach we would on ly need one launch file for the interactive teleop.

but-spanel: Florian, thanks for the tip. I'm always learning new things in ROS... I 
will fix this and let you know.

Michal

On 4.10.2012 11:32, Florian Weißhardt wrote:

> In cob_interactive_teleop/launch/interactive_teleop_safe.launch:
> 
> > @@ -0,0 +1,11 @@
> > +<?xml version="1.0"?>
> > +<launch>
> > +
> > -  <rosparam ns="cob_interactive_teleop" file="$(find cob_interactive_teleop)/config/interactive_teleop.yaml" command="load" />
> >   +
> > -  <node name="cob_interactive_teleop" pkg="cob_interactive_teleop" type="interactive_teleop" output="screen">
> > -    <remap from="/base_controller/command" to="/base_controller/command_safe" />
> 
> to prevent many launch files for just small configuration issues, 
> there are launch file arguments (see 
> http://www.ros.org/wiki/roslaunch/XML/arg) which can be used to 
> configure included launch files from parent launch files or the console.
> 
> suggestion: introduce arg "base_topic" which is by default pointing to 
> "/base_controller/command_safe" but can be overwritten by setting the 
> argument outside of the launch file.
> 
> Using this approach we would on ly need one launch file for the 
> interactive teleop.
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/ipa320/cob_command_tools/pull/15/files#r1759741.

## 

Michal Spanel
Assistant professor
Computer Graphics and Multimedia Department
Faculty of Information Technology
Brno University of Technology
Bozetechova 2, 612 66 Brno

## Czech Republic

E-mail: spanel@fit.vutbr.cz
Phone:  +420 54114-1294
Web:    http://www.fit.vutbr.cz/~spanel/

but-spanel: Florian, launch file arguments are there so this should be fixed... And I've also fixed loading of node parameters using the private node handle...

PR #25, opened 2013-06-24
Feature/rqt dashboard

abubeck:
Sorry for the dirty commits. Had to rollback some stuff.


floweisshardt:
emergency stop is missing, right?

PR #42, opened 2014-03-24
catkin_lint and install tags

fmessmer:


floweisshardt:
still waiting for changes from @ipa-nhg

floweisshardt:
tested by @ipa-nhg: works, but also non executable python files are located in 'lib' instead of the python2.7 dist directory

PR #44, opened 2014-04-09
creation of cob_teleop_cob4 with bride

ipa-fmw-ms:
Also hidden files are included so it can be opened in eclipse out of the box.


floweisshardt:
please remove .metadata files
.project and .cproject are OK to commit

floweisshardt:
visual review: OK

PR #57, opened 2014-08-26
catkin_lint'ing

fmessmer:


floweisshardt:
this is autogenerated by BRIDE. Should be fixed in the BRIDE templates, see https://github.com/ipa320/bride/issues/99

fmessmer:
I agree that we should not manually fix the CMakeLists as it is autogenerated.

However, I was to quick commiting another fix on the same branch on top of https://github.com/ipa-fxm/cob_command_tools/commit/f7fb5ce0e5d76bc44795beee9a40e9a87a863bc6, thus, I used `git revert` to undo the changes in `cob_teleop_cob4`

floweisshardt:
visual review: OK

PR #64, opened 2014-10-10
fix xml syntax

fmessmer:
please try to compile after applying changes...to avoid such bugs next time!


NadiaHG:
Visual Review: OK

PR #76, opened 2015-02-10
Bugfix ipa320/cob_robots#227

ghost:
bugfix ipa320/cob_robots#227
bugfix ipa320/cob_robots#227
problem description: robot returns success in initialization of the platform although emergency button is active and thus physically prevents performing homing procedure (motors signal communication error);
to fix that bug (in ipa320/cob_command_tools/cob_teleop): now checking success of init/recover procedures instead of just checking success of service calls;
to fix that bug (in ipa320/cob_driver/base_drive_chain): changed wrong ROS output regarding base recover service callback

### DIFF
--- a/cob_teleop/ros/src/cob_teleop_v1.cpp
+++ b/cob_teleop/ros/src/cob_teleop_v1.cpp
@@ -544,7 +544,10 @@ void TeleopCOB::joy_cb(const sensor_msgs::Joy::ConstPtr &joy_msg)
 		cob_srvs::Trigger srv = cob_srvs::Trigger();
 		if (client_init_base.call(srv))
 		{
-			ROS_INFO("Base init successfully");
+			if (!srv.response.success.data)
+				ROS_ERROR("Base init failed");
+			else
+				ROS_INFO("Base init successfully");
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


ghost:
I suggest adding a flag to the teleop_cob node for checking if base is already initialized or not; this could be useful to avoid unnecessary service calls and ROS output; I also suggest to remove redundant ROS output either in teleop_cob or base_drive_chain;

floweisshardt:
visual review: OK

cob_teleop/ros/src/cob_teleop_v1.cpp
```diff
@@ -540,27 +540,33 @@ void TeleopCOB::joy_cb(const sensor_msgs::Joy::ConstPtr &joy_msg)
...
-			ROS_INFO("Base init successfully");
+			if (!srv.response.success.data)
+				ROS_ERROR("Failed to initialize base.");
+			else
+				ROS_INFO("Base successfully initialized.");
 		}
 		else
 		{
-			ROS_ERROR("Failed to call service /base_controller/init");
+			ROS_ERROR("Failed to call service /base_controller/init.");
```
fmessmer, line 554: @ipa-mig-tk @ipa-mig 
It's the same hardcoded namespace in here, too (see [cob_robots#204](https://github.com/ipa320/cob_driver/pull/204))
Please change this!

ghost:
Done. Sending PR.

PR #84, opened 2015-03-13
remove annoying command_gui notification popups

floweisshardt:
this is really annoying because it lacks and queues up if emergency stop is activated several times, especially with wifi settings


NadiaHG:
Visual Review: OK

PR #97, opened 2015-04-29
updates from cob4-2

ipa-cob4-2:


fmessmer:
This is only an intermediate state of code and is significantly improved in #98 
Thus, this PR should be merged together with #98

PR #100, opened 2015-05-29
replace brics_actuator

fmessmer:
as discussed, we replace brics_actuator messages with ros_control-conform std_msgs/Float64MultiArray and adjusted topic names


NadiaHG:
Test cob3-2: OK

PR #106, opened 2015-08-09
Migration to package format 2

fmessmer:
I also fixed the ugly trailing whitespaces with 

```
find -not -path "*.git/*" -type f | xargs sed -i 's/[ \t]*$//'
```

Use [this link](https://github.com/ipa320/cob_command_tools/pull/106?w=1) to suppress whitespace-only changes!

@ipa-fmw @ipa-nhg @ipa-mdl @ipa-mig please check whether I considered everything as discussed


NadiaHG:
LGTM

cob_interactive_teleop/package.xml
```diff
@@ -14,20 +14,14 @@
...
+  <depend>visualization_msgs</depend>
 
-  <run_depend>geometry_msgs</run_depend>
-  <run_depend>interactive_markers</run_depend>
-  <run_depend>roscpp</run_depend>
-  <run_depend>rviz</run_depend>
-  <run_depend>std_msgs</run_depend>
-  <run_depend>tf</run_depend>
-  <run_depend>visualization_msgs</run_depend>
+  <exec_depend>rviz</exec_depend>
```
mgruhler, line 24: Would this require `exec_depend`s to `cob_bringup_sim` and `cob_navigation_global` due to the launch file `interactive_teleop_test.xml`?

cob_script_server/package.xml
```diff
@@ -13,28 +13,25 @@
...
-  <run_depend>rospy</run_depend>
-  <run_depend>rostest</run_depend>
-  <run_depend>std_msgs</run_depend>
-  <run_depend>std_srvs</run_depend>
-  <run_depend>tf</run_depend>
-  <run_depend>trajectory_msgs</run_depend>
+  <exec_depend>message_runtime</exec_depend>
+
+  <depend>actionlib_msgs</depend>
+  <depend>actionlib</depend>
```
mgruhler: Why is there the `depend` on `actionlib`? Because we build actions? The [docs](http://docs.ros.org/jade/api/catkin/html/howto/format2/building_msgs.html) say you only need `actionlib_msgs`. Or did I miss something?

cob_teleop/common/include/cob_teleop/cob_teleop_v2_common.h
```diff
@@ -2,14 +2,6 @@
...
 #include <sensor_msgs/JoyFeedbackArray.h>
 #include <geometry_msgs/Twist.h>
-#include <geometry_msgs/Twist.h>
-#include <geometry_msgs/Twist.h>
-#include <std_msgs/Float64MultiArray.h>
-#include <std_msgs/Float64MultiArray.h>
-#include <geometry_msgs/Twist.h>
-#include <std_msgs/Float64MultiArray.h>
-#include <geometry_msgs/Twist.h>
-#include <std_msgs/Float64MultiArray.h>
```
mgruhler, line 12: Can you do the same cleanup [here](https://github.com/ipa320/cob_command_tools/blob/indigo_dev/cob_teleop/ros/src/cob_teleop_v2_ros.cpp#L7-L18)?

mgruhler:
With the exception of the comments above: LGTM

cob_interactive_teleop/package.xml
fmessmer, line 24: Indeed it would, but that would introduce a dependency to the probably biggest dependency set of all cob packages...thus, I think, we should not add them...no one is going to execute this file anyway....

cob_script_server/package.xml
fmessmer: removed!

cob_teleop/common/include/cob_teleop/cob_teleop_v2_common.h
fmessmer, line 12: done

fmessmer:
I also added explicit dependencies for `boost` and removed the install-tags for headers in executable-only packages as well as the according entries in `catkin_package`....

fmessmer:
Did I use the correct Boost COMPONENTS?

mathias-luedtke:
The code does not include any boost header, so it seems to be a transitive dependency.
I do see any reference to the system component, I would just omit the component..

mathias-luedtke:
http://www.boost.org/doc/libs/1_54_0/more/getting_started/unix-variants.html#header-only-libraries lists all available components (those which need libraries to be linked)

PR #108, opened 2015-09-04
fixes type conversion in ddwrt

fmessmer:
This was done on my robot account together with @ipa-nhg 

Although I push this here, I think we want to review diagnostics for wifi as a whole!
Therefore, for now this should just be a reminder PR...


floweisshardt:
visual review: OK

PR #110, opened 2015-09-21
Deleted script_server tests from launch file

koellsch:
The script_server tests were originally included in the normal script_server launch file and were executed in any rostest which uses the script_server. You can now run the tests by executing the script_server.test file separately.


floweisshardt:
+1

PR #115, opened 2015-09-28
updates for teleop

floweisshardt:
updated during testing on cob4-2:
- disable light (due to not switching back to default breath mode)
- encapsulate say
- use deadman button to enable switch mode


floweisshardt:
superceeded by #117

Issue #112, opened 2015-09-24
[cob_teleop] Enhancements


fmessmer:
- re-add safety-overwrite to button LB in order to overwrite footprint_observer (see https://github.com/ipa320/cob_control/issues/32)
- stream `joint_group_velocity_controller/command` topic to not switch back to trajectory_controller
- finish with single zero velocity when button is released
- finish with single zero twist in twist_control mode when button is released
- reactivate light feedback for switch mode (was disabled in #117 due to not switching back to old mode, e.g. breath mode) 

@fmessmer FYI

fmessmer:
evolving from the discussion in https://github.com/ipa320/cob_robots/pull/414#issuecomment-182599919, we might want to consider renaming (some of) the parameter so that the do not carry any module name (e.g. head) in their name....
the names could be named `sss_default_target_button_module1` where the  numer than revers to the index in the `components` XMLRPC structure...

fmessmer:
not planned

PR #117, opened 2015-09-28
updates for teleop

floweisshardt:
superceeds #115 

updated during testing on cob4-2:
- disable light (due to not switching back to default breath mode)
- encapsulate say
- use deadman button to enable switch mode


cob_teleop/ros/src/cob_teleop.cpp
```diff
@@ -343,14 +345,19 @@ void CobTeleop::updateBase(){
...
 }
 
-
-void CobTeleop::joy_cb(const sensor_msgs::Joy::ConstPtr &joy_msg){
-
-  if(mode_switch_button_>=0 && mode_switch_button_<(int)joy_msg->buttons.size() && joy_msg->buttons[mode_switch_button_]==1)
+void CobTeleop::say(std::string text, bool blocking)
+{
+  cob_sound::SayGoal saygoal;
+  std::replace(text.begin(), text.end(), '_', ' ');
```
fmessmer: I'd prefer `say` to say what is passed to it.
(Thus I'd prefer to use `replace` where the saytext is composed. Even if that means some duplication)

cob_teleop/ros/src/cob_teleop.cpp
```diff
@@ -382,20 +395,32 @@ void CobTeleop::joy_cb(const sensor_msgs::Joy::ConstPtr &joy_msg){
...
-    ROS_INFO("Init button pressed");
+    ROS_INFO("Init and recover");
+    say("init and recover", true);
 
     for(std::map<std::string,XmlRpc::XmlRpcValue>::iterator p=components_.begin();p!=components_.end();++p)
     {
       std::string comp_name = p->first;
+      say(comp_name, true);
+
+      ROS_INFO("Init %s",comp_name.c_str());
```
fmessmer, line 404: Can we have some consistency, i.e. always have `ROS_INFO()` before `say()` without empty line?

fmessmer:
How can we actually fix the "light" issue - without commenting it? Any suggestion?
If we merge this without a fix for "light", it should be documented in #112

cob_teleop/ros/src/cob_teleop.cpp
floweisshardt: addressed in https://github.com/ipa-fmw/cob_command_tools/commit/23cc62115a2c9df18d07c9454f833aff178f5d01

cob_teleop/ros/src/cob_teleop.cpp
floweisshardt, line 404: addressed in https://github.com/ipa-fmw/cob_command_tools/commit/23cc62115a2c9df18d07c9454f833aff178f5d01

floweisshardt:
I fixed your remarks and added the light issue to #112.

ready for next round...

fmessmer:
would you mind taking care of "tabs vs. spaces" in cob_teleop.cpp and keyboard_publisher.cpp?

floweisshardt:
done in https://github.com/ipa-fmw/cob_command_tools/commit/de8cc45ccad4d044cda4714d6895e4ca824d673d

fmessmer:
Thanks!
Last question: Is jenkins happy with the new cob_teleop or does the build error still exist?

floweisshardt:
still exists, see http://jenkins.ros.org/job/devel-indigo-cob_command_tools/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/37/console

fmessmer:
The jenkins failure is strange....why is including SetLightMode.h possible while including ScriptAction.h fails...where is the difference between `cob_light` and `cob_script_server`?

one thing I can see is that in `cob_light` `actionlib` is defined as `<depend>` and in `cob_script_server` it only is a `<exec_depend>`...

floweisshardt:
merging this and opening issue (#118) for jenkins error

PR #122, opened 2015-11-10
fix sss error handling for light

floweisshardt:


NadiaHG:
LGTM

NadiaHG:
Test on cob4-2: OK

fmessmer:
Although this PR is minor, I don't like PRs to be accepted within the same second of being proposed!
You  don't have any chance to comment or approving...or in my case to ask a question for clarification!

cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -697,7 +697,7 @@ def compose_color(self,component_name,parameter_name):
 			full_parameter_name = self.ns_global_prefix + "/" + component_name + "/" + parameter_name
 			if not rospy.has_param(full_parameter_name):
 				rospy.logerr("parameter %s does not exist on ROS Parameter Server, aborting...",full_parameter_name)
-				return 2
+				return 2, None
```
fmessmer, line 700: can you elaborate a bit on what exactly the two values mean?
Is it error code and handle?

Is it always the case that a tuple should be returned?
It doesn't seem to be done consistently within the rest of the simple_script_server....
Could you review the this file in order to see whether the same fix should be applied for other functions, too?

floweisshardt, line 700: this fix only applies to this specific function because it is a helper function which can rise errors, but where the handle doesn't exist.
All other functions do either not rise errors or have access to `ah`

PR #123, opened 2015-11-12
fixed em stop monitor

benmaidel:
latest `sss.compose_color` returns a list of (errorcode, color_touple) instead of a simple color_touple

### DIFF
--- a/cob_monitoring/src/emergency_stop_monitor.py
+++ b/cob_monitoring/src/emergency_stop_monitor.py
@@ -132,7 +132,7 @@ class emergency_stop_monitor():
 	## set light
 	def set_light(self, color, flashing=False):
 		for component in self.light_components:
-			color_rgba = sss.compose_color(component, color)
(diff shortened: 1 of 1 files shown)


cob_monitoring/src/emergency_stop_monitor.py
```diff
@@ -132,7 +132,7 @@ def jointstate_callback(self, msg):
 	## set light
 	def set_light(self, color, flashing=False):
 		for component in self.light_components:
-			color_rgba = sss.compose_color(component, color)
+			comp, color_rgba = sss.compose_color(component, color)
```
fmessmer, line 135: If the meaning of the first member of the tuple is "error_code", why not name it `error_code`  instead of `comp`?

floweisshardt, line 135: I'll merge this and rename it to error_code

floweisshardt:
LGTM

PR #124, opened 2015-11-19
fix emergency_stop_monitor: set queue size to 1

floweisshardt:
otherwise it is infinite and will procudes longish sound outputs triggered by each callback


benmaidel:
CR: :ok:

fmessmer:
CR :ok:

NadiaHG:
CR: OK

PR #132, opened 2016-02-04
introduce enable params, styling, consistency, beautifying

### LINKED ISSUE #131: [cob_teleop] recovery of r@w base takes too long
Currently, on r@w, the recovery of the base takes very long, approx. 30s.

The problem actually is the new teleop. Even though [this answer](https://github.com/ipa320/cob_command_tools/pull/109#issuecomment-141005264) to [my question](https://github.com/ipa320/cob_command_tools/pull/109#issuecomment-140797161) whether this node works without having the `cob_sound` node started is correct in the sense that it works, there behaviour is not satisfactory. And I don't want to start `cob_sound` on a robot without speakers.

Some (hopefully correct) insights:

The problem is actually that the `say()` function [here](https://github.com/ipa320/cob_command_tools/blob/indigo_dev/cob_teleop/ros/src/cob_teleop.cpp#L357-L366) is called multiple times in the `recover` part of teleop [here](https://github.com/ipa320/cob_command_tools/blob/indigo_dev/cob_teleop/ros/src/cob_teleop.cpp#L404-L442), effectively resulting in multiple `waitForResult(5)` calls if the respective server is not there.

How should this be handled? I`d vote for a parameter to turn of this feature completely!
@ipa-srd @ipa-flg @ipa-nhg @ipa-fxm @ipa-fmw

fmessmer:
fixes #131 

How do you like this? @ipa-mig @ipa-nhg
I couldn't help but also do some styling and restructuring...

Still it needs to be tested on the robots...and we need to set the new parameters in the teleop.yaml in case sound and or light is allowed/desired...default is **both off**

### DIFF
--- a/cob_teleop/ros/src/cob_teleop.cpp
+++ b/cob_teleop/ros/src/cob_teleop.cpp
@@ -303,8 +303,8 @@ sensor_msgs::JoyFeedbackArray CobTeleop::switch_mode()
     saytext = "Cartesian mode";
   }
 
-  say(saytext, false);
   setLight(mode_);
+  say(saytext, false);
 
   LEDS_=led_mode_[mode_];
 
--- a/cob_teleop/ros/src/cob_teleop.cpp
+++ b/cob_teleop/ros/src/cob_teleop.cpp
@@ -77,22 +77,23 @@ class CobTeleop
 {
 public:
 
-  struct component_config{
+  struct component_config
+  {
     std::string key;
     std::string twist_topic_name;
     std::string vel_group_topic_name;
     std::string sss_default_target;
     std::vector<double> joint_velocity;
+    std::vector<double> twist_max_vel; //max_vx_, max_vy_, max_vz_, max_rotx_, max_roty_, max_rotz_
+    std::vector<double> twist_max_acc; //max_ax_, max_ay_, max_arotz_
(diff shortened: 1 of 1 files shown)


cob_teleop/ros/src/cob_teleop.cpp
```diff
@@ -262,59 +282,55 @@ sensor_msgs::JoyFeedbackArray CobTeleop::switch_mode(){
...
-  color.r = 0;
-  color.g = 1;
-  color.b = 0;
-  color.a = 1;
-  light.pulses = mode_;
-  light.color = color;
-  lightgoal.mode = light;
-//  setlightclient_->sendGoal(lightgoal);
   say(saytext, false);
+  setLight(mode_);
```
mgruhler, line 306: This is more an aesthetical issue:
Do we want to switch the light **after** the robot has said what it is about to say (which could take a few seconds), or **before**?

I'd say before, but this is just personal preference ;-)

fmessmer, line 306: true, I was just thinking about conistently having always script, sound, light in that order....but here, I did not think about the consequences...light first makes definitely more sense here! I'll change it!

mgruhler:
CR: ok
tested on raw3-3: ok

NadiaHG:
FR: OK

PR #133, opened 2016-02-12
rework sss timeouts

floweisshardt:


NadiaHG:
FR: OK
Test on cob4-2: OK

fmessmer:
There are other calls to `wait_for_server` still having the 5 seconds duration...I'd say we should reduce them all to the same timeout...or is there a reason we should wait longer for any of the other clients?

fmessmer:
THX

PR #137, opened 2016-02-22
reduce terminal output

floweisshardt:


fmessmer:
FR ok....was annoying indeed....

PR #138, opened 2016-02-24
stop charging mode if no more power_state msgs received

benmaidel:


floweisshardt:
FR: OK, I assume you tested on cob4-1

benmaidel:
that's right. Tested it on cob4-1

fmessmer:
So what does this PR solve?
The way I read it is, it stops flashing the led in case no PowerState message is received for a while, ritght?
It does **not** solve the problem, that it stops flashing in case the `bms_driver` is not running as the `power_state_aggregator` keeps publishing in it's own `spin()`-Thread...

fmessmer:
as explained by @ipa-fmw in https://github.com/ipa320/cob_power_management/pull/19#issuecomment-196398743, my second issue is fixed in https://github.com/ipa320/cob_power_management/pull/19 ...thus merging

PR #141, opened 2016-03-05
Feature script server feedback in command gui

### LINKED ISSUE #139: [cob_script_server] display (Trigger-)Response within cob_command_gui
Currently you do not get any notification if e.g. `init` or `recover` calls fail!

@fmessmer FYI

fmessmer:
fixes #139 

calling the sss.command calls in a dedicated function (thanks @ipa-bnm) that evaluates the action_handle and displays the according messages in a MessageDialog (in case the action_handles returns a failure)
required quite a bit consistency tuning in the actual script_server code, too

Looks like this:
knoeppkes_error_dialog

Please can I get at least 3 reviews to ensure I did not mess up any python syntax in the cob_script_server?
@ipa-fmw, @ipa-nhg, @ipa-mdl, @ipa-bnm, @ipa-mig

### DIFF
--- a/cob_command_gui/src/knoeppkes.py
+++ b/cob_command_gui/src/knoeppkes.py
@@ -102,7 +102,24 @@ def start(func, args):
         largs.append("diff")
     #print "Args", tuple(largs)
     #print "func ", func
-    thread.start_new_thread(func,tuple(largs))
+    #thread.start_new_thread(func,tuple(largs))              # exits silently without evaluating result
+    thread.start_new_thread(call_thread,(func,tuple(largs))) # calls func(args) and shows result in a MessageDialog
+
+def call_thread(func,args):
+  result = func(*args)
+  if not result.success:    # action_handle returns with failure
+    result_dialog = gtk.MessageDialog(None, 0, gtk.MESSAGE_ERROR,
+                                               gtk.BUTTONS_OK,
+                                               "Executing " + func.__name__ + "(" + args[0] + ") failed!")
+    gtk.gdk.threads_enter()
+    result_dialog.format_secondary_text(result.message)
+    gtk.gdk.threads_leave()
+    gtk.gdk.threads_enter()
+    result_dialog.run()
+    gtk.gdk.threads_leave()
+    gtk.gdk.threads_enter()
+    result_dialog.destroy()
+    gtk.gdk.threads_leave()
 
 def startGTK(widget, data):
   data()
@@ -169,19 +186,19 @@ class GtkGeneralPanel(gtk.Frame):
 
   def stop_all(self,component_names):
     for component_name in component_names:
-      self.sss.stop(component_name,blocking=False)
+      start(self.sss.stop,(component_name, True)) # services always blocking
 
   def init_all(self,component_names):
     for component_name in component_names:
-      self.sss.init(component_name,False)
+      start(self.sss.init,(component_name, True)) # services always blocking
 
   def recover_all(self,component_names):
     for component_name in component_names:
-      self.sss.recover(component_name,False)
+      start(self.sss.recover,(component_name, True)) # services always blocking
 
   def halt_all(self,component_names):
     for component_name in component_names:
-      self.sss.halt(component_name,False)
+      start(self.sss.halt,(component_name, True)) # services always blocking
 
 
   def setEMStop(self, em):
@@ -284,10 +301,10 @@ class Knoeppkes():
     gtk.gdk.threads_leave()
 
 def signal_handler(signal, frame):
-#  print 'You pressed Ctrl+C!'
+  #print 'You pressed Ctrl+C!'
   gtk.main_quit()
 
 if __name__ == "__main__":
-	signal.signal(signal.SIGINT, signal_handler)
-	app = Knoeppkes()
+  signal.signal(signal.SIGINT, signal_handler)
+  app = Knoeppkes()
 
--- a/cob_command_gui/src/knoeppkes.py
+++ b/cob_command_gui/src/knoeppkes.py
@@ -81,7 +81,7 @@ initialized = False
 #gtk.gdk.threads_init()
 
 ## Executes a button click in a new thread
-def start(func, args):
+def start(func, args, dialog=False):
   global planning_enabled
   global base_diff_enabled
   global confirm_commands_enabled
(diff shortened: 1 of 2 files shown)


fmessmer:
Still an issue is that command_gui calls `move` commands with `Blocking=False`! In that case, the `action_handle` is never `set_succeeded`! Which is why a pop-up will pop up saying that the action failed....while the arm is still moving!

fmessmer:
I'm still waiting for feedback/review of this PR!
If the PopUp is not desired....the consistency changes within cob_script_server are still worth merging!

cob_command_gui/src/knoeppkes.py
```diff
@@ -102,7 +102,24 @@ def start(func, args):
...
+                                               "Executing " + func.__name__ + "(" + args[0] + ") failed!")
+    gtk.gdk.threads_enter()
+    result_dialog.format_secondary_text(result.message)
+    gtk.gdk.threads_leave()
+    gtk.gdk.threads_enter()
+    result_dialog.run()
+    gtk.gdk.threads_leave()
+    gtk.gdk.threads_enter()
+    result_dialog.destroy()
+    gtk.gdk.threads_leave()
```
mgruhler, line 124: I'm not familiar with the `gtk` thread API... Why do you need to `leave()` and `enter` after every change to `result_dialog`?
My intuition would be that this is only required before the `destroy()` if at all, but I might be mistaken ;-)

cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -266,46 +269,45 @@ def trigger(self,component_name,service_name,blocking=True, planning=False):
 		rospy.loginfo("<<%s>> <<%s>>", service_name, component_name)
 		parameter_name = self.ns_global_prefix + "/" + component_name + "/service_ns"
 		if not rospy.has_param(parameter_name):
-				rospy.logerr("parameter %s does not exist on ROS Parameter Server, aborting...",parameter_name)
-				ah.set_failed(2)
+				message = "Parameter " + parameter_name + " does not exist on ROS Parameter Server, aborting..."
+				rospy.logerr(message)
+				ah.set_failed(2, message)
```
mgruhler: indentation by 2 tabs?

cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -355,26 +359,29 @@ def move_base(self,component_name,parameter_name,blocking, mode):
 
 		# check trajectory parameters
 		if not type(param) is list: # check outer list
-				rospy.logerr("no valid parameter for %s: not a list, aborting...",component_name)
-				print "parameter is:",param
-				ah.set_failed(3)
-				return ah
+			message = "No valid parameter for " + component_name + ": not a list, aborting..."
+			rospy.logerr(message)
+			print "parameter is:",param
```
mgruhler, line 362: I know this has been like this before, but shouldn't this rather be a `logwarn` (or `loginfo`)?

cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -862,27 +886,31 @@ def say(self,component_name,parameter_name,blocking=True):
...
 
 		# check parameters
 		if not type(param) is list: # check list
-				rospy.logerr("no valid parameter for %s: not a list, aborting...",component_name)
-				print "parameter is:",param
-				ah.set_failed(3)
-				return ah
+			message = "No valid parameter for " + component_name + ": not a list, aborting..."
+			rospy.logerr(message)
+			print "parameter is:",param
```
mgruhler, line 886: again a `print`... there are some more in here as well, which you didn't touch. Change, leave or set up an issue? I don't mind...

NadiaHG:
Could be possible add the pop-up feature to the check button list of the command_gui? To enable or disable the notifications? 

P.S. Consider the comment, only if it doesn't require too much effort.

cob_command_gui/src/knoeppkes.py
fmessmer, line 124: I'm not too familar either, but I had problems when **several** pop-ups were opend (e.g. when pressing "INIT ALL") and this solution was proposed in a QA forum...

cob_script_server/src/simple_script_server/simple_script_server.py
fmessmer, line 362: `logerr` seems fine as it results in **not moving** and `failure` within the `action_handle`

fmessmer:
Has anyone tested this branch and seen/played with the pop-ups? Or was it just "Visual Review"?
@ipa-mig @ipa-nhg

I'm asking for "Usage Experience" here...as I'm currently tending to remove the pop-ups again...or comment them out and use the old behavior....then this PR would rather be about `action_handle` consistency within `cob_script_server`
(@ipa-nhg allowing to disable the pop-up wouldn't make too much sense....it would result in always disabling rather than fixing the problem :wink:)

cob_script_server/src/simple_script_server/simple_script_server.py
fmessmer, line 886: Was like this before...thus keep!

cob_script_server/src/simple_script_server/simple_script_server.py
mgruhler, line 362: sorry, should have been more specific. Was talking about the print...

mgruhler:
Just visual...

NadiaHG:
I only tested it on simulation, but the pop-up appears always after each command immediately although the component is moving, because we haven't got proper diagnostics (?) or I have to check other branch/repositories? For this reason I proposed the "disable notifications" option

fmessmer:
@ipa-nhg: answer to your question :point_right: https://github.com/ipa320/cob_command_tools/pull/141#issuecomment-194323217

fmessmer:
I changed the behavior to only show the pop-up for `init` and `recover` (as this was the original intention of the issue)

Also fixed indentation...

floweisshardt:
I don't like the approach of introducing pop ups. talking to @ipa-fxm we'll keep the consistency changes and revert the pop-up feedback

floweisshardt:
DO NOT MERGE, this changed sss behaviour (all service calls are blocking, there's no chance of calling service in a non-blocking way)

fmessmer:
closing in favor of #142

PR #161, opened 2016-10-12
hz monitor: make sampling rate configurable and add warning

floweisshardt:
will not change default behaviour (1Hz)


mgruhler (approved):
FR: ok

PR #164, opened 2016-12-08
implemented service_relay

mathias-luedtke:


service_tools/scripts/service_relay
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
```
fmessmer, line 1: why not have the ```.py```-suffix?

service_tools/scripts/service_relay
```diff
@@ -0,0 +1,21 @@
...
+if __name__ == "__main__":
+    argv = rospy.myargv()
+    if len(argv) != 3:
+        rospy.logerr("service_relay IN OUT")
+
+    rospy.init_node("service_relay")
+
+    local_name, remote_name = argv[1], argv[2]
+
+    rospy.wait_for_service(remote_name)
```
fmessmer, line 15: do we need some logoutput here in case the service is not available? timeout?

service_tools/package.xml
```diff
@@ -0,0 +1,18 @@
...
+
+  <maintainer email="fmw@ipa.fhg.de">Florian Weisshardt</maintainer>
+  <author email="mathias.luedtke@ipa.fraunhofer.de">Mathias Lüdtke</author>
+
+  <license>LGPL</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <exec_depend>rosservice</exec_depend>
+  <exec_depend>rospy</exec_depend>
```
fmessmer, line 15: pedantic: alphabetic order

service_tools/package.xml
```diff
@@ -0,0 +1,18 @@
+<package format="2">
+  <name>service_tools</name>
+  <version>0.0.0</version>
```
fmessmer: could you set the version to ```0.6.4``` - otherwise we'll have some trouble on the next release

fmessmer (changes requested):
there are a few things commented inline
(I'm not yet too familar with the new github-functionality

service_tools/scripts/service_relay
mathias-luedtke: "To keep the user API clean, executable script names generally do not include a .py suffix."
(http://docs.ros.org/jade/api/catkin/html/howto/format2/installing_python.html)

service_tools/scripts/service_relay
mathias-luedtke: For our use case we don't have a default timeout. (It depends on someone to press the init button..)

mathias-luedtke:
added version and sorted deoendencies

fmessmer (approved):

PR #169, opened 2017-01-19
rospy.sleep exception handling

fmessmer:
some `sleep()` exception handling which prevents a lot of terminal output on "STRG+C"

### DIFF
--- a/cob_dashboard/scripts/cob_dashboard_aggregator.py
+++ b/cob_dashboard/scripts/cob_dashboard_aggregator.py
@@ -75,4 +75,7 @@ if __name__ == "__main__":
   r = rospy.Rate(1)
   while not rospy.is_shutdown():
     da.publish()
-    r.sleep()
+    try:
+        r.sleep()
+    except rospy.exceptions.ROSInterruptException as e:
+        pass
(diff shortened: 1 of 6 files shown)


cob_dashboard/scripts/cob_dashboard_aggregator.py
```diff
@@ -75,4 +75,8 @@ def publish(self):
   r = rospy.Rate(1)
   while not rospy.is_shutdown():
     da.publish()
-    r.sleep()
+    try:
+        r.sleep()
+    except rospy.exceptions.ROSInterruptException as e:
+        #print "ROSInterruptException", e
```
mgruhler: why not remove this line (same below)?

fmessmer: copied it from some other script which had it...

floweisshardt (approved):

cob_dashboard/scripts/cob_dashboard_aggregator.py
fmessmer: convinced: it's cleaner without

fmessmer:
Added proper `from __future__ import print_function` as suggested in https://github.com/ipa320/cob_control/issues/129#issuecomment-274029626 and [python_2_and_3_compatible_code](http://wiki.ros.org/python_2_and_3_compatible_code#Print_is_now_a_function)

NadiaHG (approved):

PR #170, opened 2017-01-20
generic topic status monitor based on diagnostic updater

fmessmer:
This is a node, I wrote once for another project that can be used to acquire diagnostics for any given topic.
It's superior to `hz_monitor` as it shows you a more detailed info on the topic status, it can also monitor TimeStamps in the header of the messages...and cpp is likely more efficient :wink:

PR #173, opened 2017-01-26
remove useless code

souravran:


fmessmer:
how is your experience with https://github.com/ipa320/cob_command_tools/pull/170

cob_monitoring/src/hz_monitor.py
```diff
@@ -142,9 +141,8 @@ def publish_diagnostics(self, rt_HZ_store = []):
                     publishing_rate_error = True
                     consolidated_error_messages["publishing rate is too high for topics"].append(topic)
                 else:
-                    if not publishing_rate_error:
-                        hz_status.level = DiagnosticStatus.OK
-                        hz_status.message = 'all publishing rates are ok'
+                    hz_status.level = DiagnosticStatus.OK
+                    hz_status.message = 'all publishing rates are ok'
```
fmessmer: I guess the part
`if not publishing_rate_error:`
is not useless in the case where one hz_monitor instance monitors the publish rate of several topics...because with your changes, it would set `DiagnosticStatus.OK` although maybe some other topics are "too low" or "too high"...

fmessmer (changes requested):
**Do not merge**

fmessmer (approved):

PR #175, opened 2017-01-30
add auto_recover to new cob_helper_tools pkg

fmessmer:
@ipa-jba @ipa-mig @ipa-fmw 
How do you like this?
Related to https://github.com/ipa320/ipa_navigation_common/pull/90

I took msh-version and made it configurable via rosparam. You can use it like - **Usage updated**:
```
<?xml version="1.0" ?>
<launch>
  <node pkg="cob_helper_tools" type="auto_recover.py" name="auto_recover" output="screen">
    <rosparam param="components">[base, torso]</rosparam>
  </node>
</launch>
```

Please comment in case something from the navigation-version is not yet covered by this script!
Also, we might want to consider other scripts from msh to this new package for more prominent usage...e.g. auto_init, auto_localisation,...anything else?
The name is pretty general, so we might even put some fake nodes in here such as the ones from soon-to-be-removed `cob_controller_configuration_gazebo`

### DIFF
--- a/cob_helper_tools/scripts/auto_recover.py
+++ b/cob_helper_tools/scripts/auto_recover.py
@@ -1,4 +1,30 @@
 #!/usr/bin/python
+#################################################################
+##\file
+#
+# \note
+#   Copyright (c) Felix Messmer \n
+#   Fraunhofer Institute for Manufacturing Engineering
+#   and Automation (IPA) \n
+#
+#   All rights reserved. \n\n
+#
+#################################################################
+#
+# \note
+#   Repository name: cob_command_tools
+# \note
+#   ROS package name: cob_helper_tools
+#
+# \author
+#   Author: Felix Messmer
+#
+# \date Date of creation: January 2017
+#
+# \brief
+#   A script to automatically recover hardware after e-stop and HW failure
+#
+#################################################################
 
 import rospy
 
--- a/cob_helper_tools/scripts/auto_recover.py
+++ b/cob_helper_tools/scripts/auto_recover.py
@@ -1,0 +1,45 @@
+#!/usr/bin/python
+
+import rospy
+
+from cob_msgs.msg import EmergencyStopState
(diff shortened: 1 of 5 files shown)


cob_helper_tools/scripts/auto_recover.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/python
+
```
mgruhler: maybe we should start again to put [the headers](https://github.com/ipa320/setup/tree/master/templates) in our sourcefiles.

mgruhler (commented):
looks good to me.

What is the difference between the em state recover and the diagnostics recover?

I figure that the former happens automatically after an em stop has been released. But what about the latter? When does this occur?

fmessmer:
I guess the second comes from temporary, recoverable hardware failures... :wink:

cob_helper_tools/scripts/auto_recover.py
floweisshardt: +1

floweisshardt (approved):

floweisshardt:
anoter script which would fit in here is `auto_init.py`

ipa-flg:
@ipa-fxm :top:

mgruhler (changes requested):
enforcing license headers :-)

ipa-flg:
@ipa-fxm 
Above usage example was wrong. Use the following:
```
<!-- auto init base -->
<node pkg="cob_helper_tools" type="auto_init.py" name="auto_init" output="screen">
  <rosparam param="components">[base]</rosparam>
</node> 
```

ipa-flg:
@ipa-fxm tested auto_init and auto_recover on raw3-3, both working fine :+1:

fmessmer:
@ipa-mig 
Done license header...please approve and merge!

mgruhler (approved):

PR #186, opened 2017-03-21
sss action trigger

### LINKED ISSUE #185: [script_server] allow to call actions from Knoepkes
let's start with simple `actionlib/TestAction` in order to simplify scenario development...
we might add more ActionTypes (or an even more generic solution) later

fmessmer:
@ipa-bnm 
extended sss by a `trigger_action` function that can be used to trigger an `actionlib/TestAction` action
fix #185 

simply run a `rosrun actionlib axserver.py /behavior_action/test_action`
and add the following to the `command_gui_buttons.yaml`:
```
group30:
  group_name: behavior_action
  component_name: behavior_action
  buttons: [[test_action,trigger_action,test_action]]
```

### DIFF
--- a/cob_script_server/src/simple_script_server/simple_script_server.py
+++ b/cob_script_server/src/simple_script_server/simple_script_server.py
@@ -88,6 +88,8 @@ from cob_sound.msg import *
 from cob_script_server.msg import *
 from cob_light.msg import LightMode, LightModes, SetLightModeGoal, SetLightModeAction
 from cob_mimic.msg import SetMimicGoal, SetMimicAction
+from actionlib.msg import TestAction, TestGoal
+from actionlib import GoalStatus
 
 graph=""
 graph_wait_list=[]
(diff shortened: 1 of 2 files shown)


cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -314,6 +316,57 @@ def trigger(self,component_name,service_name,blocking=True):
...
+
+		rospy.loginfo("<<%s>> <<%s>>", action_name, component_name)
+		#parameter_name = self.ns_global_prefix + "/" + component_name + "/service_ns"
+		#if not rospy.has_param(parameter_name):
+			#message = "Parameter " + parameter_name + " does not exist on ROS Parameter Server, aborting..."
+			#rospy.logerr(message)
+			#ah.set_failed(2, message)
+			#return ah
+		#service_ns_name = rospy.get_param(parameter_name)
+		action_full_name = "/" + component_name + "/" + action_name
```
floweisshardt, line 341: instead of (semi-)hardcoded `/component_name/action_name` we should use a lockup similar to trigger service 
```
parameter_name = self.ns_global_prefix + "/" + component_name + "/action_ns"
```
This way we can configure the action namespace in a generic way in e.g. cob_default_robot_config

fmessmer: I kind of did not use the parameter approach on purpose as imho this mechanism is outdated and now only produces quite some (avoidable) configuration overhead. Maybe we can get rid of this parameter approach for good after all...but I'm happy to discuss this further

benmaidel (approved):

floweisshardt (approved):

PR #187, opened 2017-04-10
retry init on failure

fmessmer:
fix https://github.com/ipa320/msh/issues/693


floweisshardt (approved):
tested on cob4-paul-stuttgart: OK

PR #193, opened 2017-06-30
Feature/generic throttle

MattiaRacca:


fmessmer:
Hmm, can you explain the benefit of this implementation?
How is it different from http://wiki.ros.org/topic_tools/throttle?

MattiaRacca:
Sure. First, topic_tools/throttle instantiate a node for each throttled topic while this implementation can throttle several topics with only one node. Each topic has its own subscriber/publisher pair and its own pair of thread (one running the publisher at the requested rate and one for the subscriber callback).
All parameters of the generic throttle are defined through ros parameters. An example of .yaml file follows:
```
generic_throttle:
  namespace: throttled_topics
  delay: 5

  topic_framerate:
    - [/torso_cam3d_left/rgb/image_raw, 1]
    - [/torso_cam3d_right/rgb/image_raw, 1]
    - [/base/joint_states, 1]
```
Second, the standard behaviour of the generic_throttle is to latch for a certain "delay" the last message if no new message become available. After "delay" seconds, the throttle will stop publishing the old message and wait for new messages to become available. This degree of control is not possible with topic_tools/throttle that is implemented with a simple dequeue. Although now it is the default (required by  @ipa-fmw ), the latched behaviour could be made optional.

Third, although it is not in this pull request, I have ready a version of the throttle that, for topics that publish sensor_msgs/Image, can adjusts image resolution before publishing.

generic_throttle/package.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<package>
```
fmessmer: please use package format2

generic_throttle/package.xml
```diff
@@ -0,0 +1,25 @@
...
+  <build_depend>rospy</build_depend>
+  <build_depend>roslib</build_depend>
+  <build_depend>std_msgs</build_depend>
+  <build_depend>sensor_msgs</build_depend>
+  <build_depend>visualization_msgs</build_depend>
+  <run_depend>rospy</run_depend>
+  <run_depend>roslib</run_depend>
+  <run_depend>std_msgs</run_depend>
+  <run_depend>sensor_msgs</run_depend>
+  <run_depend>visualization_msgs</run_depend>
```
fmessmer: some of the listed dependencies are not used....please remove them

generic_throttle/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(generic_throttle)
+
+find_package(catkin REQUIRED COMPONENTS
+  rospy
+  roslib
+  std_msgs
+  sensor_msgs
+  visualization_msgs
```
fmessmer: entries only needed for c++ code...not for python....remove the ROS packages...just keep catkin...

fmessmer (changes requested):
some comments

fmessmer:
Thanks for the explanation

 > Second, the standard behaviour of the generic_throttle is to latch for a certain "delay" the last message if no new message become available. After "delay" seconds, the throttle will stop publishing the old message and wait for new messages to become available. This degree of control is not possible with topic_tools/throttle that is implemented with a simple dequeue. Although now it is the default (required by @ipa-fmw ), the latched behaviour could be made optional.

If this is meant to provide the latched behavior, we should make it a latched behavior...i.e. if latched is desired, then it should hold the last message forever...

You could also add a lazy behavior that only publishes the throttled topic in case there is a subscriber using the [getNumSubscribers](http://docs.ros.org/api/roscpp/html/classros_1_1Publisher.html#ad56c8e954020290dcb87cd6bbd330a2b) function

generic_throttle/package.xml
```diff
@@ -0,0 +1,16 @@
...
+  <description>This package provides a throttle for ROS topics</description>
+
+  <license>LGPL</license>
+  
+  <maintainer email="mattia.racca@aalto.fi">Mattia Racca</maintainer>
+  <author email="mattia.racca@aalto.fi">Mattia Racca</author>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <build_depend>rospy</build_depend>
```
fmessmer: remove ```<build_depend>rospy</build_depend>```

fmessmer (changes requested):

generic_throttle/package.xml
MattiaRacca: Ouch, I missed that one...

MattiaRacca:
> If this is meant to provide the latched behavior, we should make it a latched behavior...i.e. if latched is desired, then it should hold the last message forever...

We can do that by default the throttle latches the last message forever and, if the delay parameter is given, it behaves like the current code (drop after "delay" seconds). How does it sounds?

> You could also add a lazy behavior that only publishes the throttled topic in case there is a subscriber using the getNumSubscribers function

Sure, that's easy. I'll do tomorrow.

fmessmer:
> We can do that by default the throttle latches the last message forever and, if the delay parameter is given, it behaves like the current code (drop after "delay" seconds). How does it sounds?

I don't see a use-case for `delay`....
I'd just go with a boolean parameter `latched` which stores the last message forever or not at all...

generic_throttle/README.md
```diff
@@ -0,0 +1,20 @@
+# generic_throttle
+
+generic_throttle is package offering a Python implementation of a throttle for topics.
+
+Class GenericThrottle runs a ROS node which subscribes to a list of topics and
```
fmessmer: ~~Am I missing something: there is no node that actually uses the GenericThrottle python module...
Are you going to add one? Where/How are you currently using this package/node?~~

Found the ```rospy.init_node```.
But still there is now ```__main__```/executable, right?

I'd rather have an executable python ```generic_throttle_node.py``` which makes use of ```generic_throttle.GenericThrottle``` class but does all the "node initialization" and spinning...

LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
fmessmer: Do you have a specific interest in adding the LICENSE explicitly?
We normally do not do this for our packages...

generic_throttle/README.md
```diff
@@ -0,0 +1,20 @@
...
+publishes them again at a desired rate. List of topics and desired rate are
+specified through ROS parameters.
+
+By default, the throttle operates in a latched mode: if no new message are available
+from the input topics, it will continue to publish the last available message.
+However, after a certain time (to be defined in parameter /generic_throttle/delay
+in seconds) it will throw a warning and stop publishing the old message. As soon as
+new messages become available, the throttle will start to publish again.
```
fmessmer: Please update the README and describe an example yaml layout, i.e. mandatory and optional parameters

generic_throttle/src/generic_throttle/generic_throttle.py
```diff
@@ -0,0 +1,151 @@
...
+class GenericThrottle:
+    def __init__(self):
+        self.topic_dictionary = None
+        self.parameters = {'namespace': None,
+                           'topic_framerate': None,
+                           'latched': False,
+                           'lazy': False}
+
+        rospy.init_node('generic_throttle')
+        rospy.on_shutdown(self._shutdown)
```
fmessmer: see above: please move into an executable ```generic_throttle_node.py``` script

generic_throttle/src/generic_throttle/generic_throttle.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # Read mandatory /generic_throttle/* parameters from server
+        parameter_list = ['namespace', 'topic_framerate']
+        for parameter_name in parameter_list:
+            parameter_string = '/generic_throttle/' + parameter_name
```
fmessmer: do not assume absolute namespace....just look for ```parameter_name``` under the node's namespace, i.e. ```parameter_string = parameter_name``` (of course then ```parameter_string``` is obsolete!)

generic_throttle/src/generic_throttle/generic_throttle.py
```diff
@@ -0,0 +1,151 @@
...
+                self.parameters[parameter_name] = \
+                    rospy.get_param(parameter_string)
+            else:
+                rospy.logerr('Parameter ' + parameter_string + ' not available')
+                exit(5)
+
+        # Read optional /generic_throttle/* parameters from server
+        parameter_list = ['latched', 'lazy']
+        for parameter_name in parameter_list:
+            parameter_string = '/generic_throttle/' + parameter_name
```
fmessmer: do not assume absolute namespace....just look for ```parameter_name``` under the node's namespace, i.e. ```parameter_string = parameter_name``` (of course then ```parameter_string``` is obsolete!)

generic_throttle/src/generic_throttle/generic_throttle.py
```diff
@@ -0,0 +1,151 @@
...
+        self.parameters = {'namespace': None,
+                           'topic_framerate': None,
+                           'latched': False,
+                           'lazy': False}
+
+        rospy.init_node('generic_throttle')
+        rospy.on_shutdown(self._shutdown)
+
+        # Read mandatory /generic_throttle/* parameters from server
+        parameter_list = ['namespace', 'topic_framerate']
```
fmessmer: I guess ```topic_name``` is better than ```namespace```

generic_throttle/src/generic_throttle/generic_throttle.py
```diff
@@ -0,0 +1,151 @@
...
+                        self.parameters['topic_framerate'])
+
+        if(not(size_flag)):
+            rospy.logerr('Parameter /generic_throttle/framerate must be'
+                         ' a list of size 2 lists')
+            exit(10)
+
+        # Populate the dictionary for the ros topic throttling
+        self._populate_dictionary()
+        rospy.spin()
```
fmessmer: move the spinning to the executable node

generic_throttle/src/generic_throttle/generic_throttle.py
```diff
@@ -0,0 +1,151 @@
...
+            if topic_info[0] is None:
+                rospy.logwarn('Cannot find topic ' + topic_id)
+
+                self.topic_dictionary[topic_id]['lock'].release_lock()
+                return
+            else:
+                # Create publisher
+                self.topic_dictionary[topic_id]['publisher'] = \
+                    rospy.Publisher(self.parameters['namespace'] + topic_id,
+                                    topic_info[0], queue_size=1)
```
fmessmer: simply use the ```latched``` option of the Publisher constructor (see http://docs.ros.org/api/rospy/html/rospy.topics.Publisher-class.html#__init__) for achieving the latched behavior....no need to save the last message yourself....

fmessmer (changes requested):
more detailed review...

I also have a suggestion for the parameter layout (this is pseudo-yaml using a list of topics with a dictionary for each topic item :wink:)
```
topics:
 - topic_name_1: {latched: False, lazy: True, topic_rate: 1.0}
 - topic_name_2: {latched: True, lazy: False, topic_rate: 42.0}
```

This way, you could specify ```latched, lazy, and frame_rate``` for each topic individually...you could then also assume all to be mandatory

I'd remove the ```namespace``` parameter and simply concatenate topic_name with ```_throttled``` suffix

LICENSE
MattiaRacca: Actually there is no need and this should not be there, I'll remove it.

MattiaRacca:
Thanks, working on it.
> I'd remove the namespace parameter and simply concatenate topic_name with _throttled suffix
What if 2 throttles subscribe to the same topic? they would try to publish on the same _throttled topic, not respecting the specifications on rate.

I would actually put in front of the original topic name the namespace of the throttle node
```
/throttle1/throttle <--- this is the node for the first throttle in its namespace
/throttle1/camera/image_raw <--- throttled topic by the first throttle
/throttle2/throttle <--- second throttle in its namespace
/throttle2/camera/image_raw <--- throttled topic by the second throttle
```

fmessmer:
From the subscriber-perspective, it is counter intuitive to have the node name of the throttle-ing node in the topic name...the subscriber actually should not care about who is going to provide him with the throttled topic...he is just interested in being able to subscribe to the topic he is interested in but with a throttled frequency...thus topic_name + ```_throttled```-suffix....

generic_throttle/src/generic_throttle/generic_throttle.py
MattiaRacca: If I use the `latch = True`, the publisher will yes keep the last message latched but a new subscriber to the throttled topic will receive **only once** the message and not at the rate specified in the throttle parameters.
Could this be fine?

generic_throttle/package.xml
```diff
@@ -0,0 +1,15 @@
+<package format="2">
+  <name>generic_throttle</name>
+  <version>0.1.5</version>
+  <description>This package provides a throttle for ROS topics</description>
+
+  <license>LGPL</license>
+  
+  <maintainer email="mattia.racca@aalto.fi">Mattia Racca</maintainer>
```
fmessmer: I guess we should add someone from IPA as a maintainer, too, as it lives in the ipa320 organization...

generic_throttle/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(generic_throttle)
+
+find_package(catkin REQUIRED COMPONENTS)
+
+catkin_python_setup()
+
+catkin_package()
```
fmessmer, line 8: please add install tags for the executable node

generic_throttle/src/generic_throttle_node.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/python
```
fmessmer: I think the generic_throttle_node.py should node be within the src folder - which is reserved for the python module...
put it under generic_throttle/scripts

fmessmer (changes requested):
few minor requests....and then I would like to ask @ipa-fmw to verify the behavior for the intended use-case....

generic_throttle/package.xml
MattiaRacca: I agree. @ipa-fxm  or @ipa-fmw?

floweisshardt (approved):

floweisshardt:
@MattiaRacca adressed all comments from @ipa-fxm, so good to merge

PR #194, opened 2017-07-10
Feature/resolution throttle

MattiaRacca:
Addition to generic throttle (#193):
For sensor_msgs/Image you can define a resolution reduction and this will be applied before throttling. See README.md for more details.

### DIFF
--- a/generic_throttle/README.md
+++ b/generic_throttle/README.md
@@ -12,17 +12,19 @@ test_throttle:          # Throttle namespace
   topics:               # List of topics to be throttled
     - /topic1: {latched: False, lazy: True, topic_rate: 1.0}
     - /topic2: {latched: True, lazy: False, topic_rate: 20.0}
+    - /image_topic: {latched: False, lazy: True, topic_rate: 20.0, resolution: 2}
 ```
 For each topic, 3 parameters must be specified:
 - `latched`: if `True`, the publisher of the throttled topic acts as latched (see  http://docs.ros.org/api/rospy/html/rospy.topics.Publisher-class.html)
 - `lazy`: if `True`, the publisher of the throttled topics will check if the number of subscribers and if there are no, it will not publish the message
 - `topic_rate`: desired rate (in Hz) for the throttled topic.
 
+Optional parameter:
+- `resolution`: it is meant for topics of type sensor_msgs/Image. If available, the throttle will reduce the resolution of the image by the specified factor. For example, if `resolution: 2`, width and height will be halved.
+
 The throttle will publish on topics with name `/topic_throttled` (it appends `_throttled` to the original name).
 When running an instance of `generic_throttle_node`, you must define the throttle namespace to let the node discover the required parameters. An example from .launch file:
 ```
 <node pkg="generic_throttle" type="generic_throttle_node.py" name="throttle" ns="test_throttle" />
 ```
-# TODO
-- for `sensor_msgs/Image` enable reduction of the resolution of the throttled topic
-- for `sensor_msgs/PointCloud2` enable reduction (VoxelGrid sampling) of points in the throttled topic
+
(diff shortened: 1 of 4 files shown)


generic_throttle/README.md
```diff
@@ -12,17 +12,19 @@ test_throttle:          # Throttle namespace
   topics:               # List of topics to be throttled
     - /topic1: {latched: False, lazy: True, topic_rate: 1.0}
     - /topic2: {latched: True, lazy: False, topic_rate: 20.0}
+    - /image_topic: {latched: False, lazy: True, topic_rate: 20.0, resolution: 2}
```
floweisshardt: I would call this ```resolution_factor```

generic_throttle/README.md
```diff
@@ -12,17 +12,19 @@ test_throttle:          # Throttle namespace
...
 ```
 For each topic, 3 parameters must be specified:
 - `latched`: if `True`, the publisher of the throttled topic acts as latched (see  http://docs.ros.org/api/rospy/html/rospy.topics.Publisher-class.html)
 - `lazy`: if `True`, the publisher of the throttled topics will check if the number of subscribers and if there are no, it will not publish the message
 - `topic_rate`: desired rate (in Hz) for the throttled topic.
 
+Optional parameter:
+- `resolution`: it is meant for topics of type sensor_msgs/Image. If available, the throttle will reduce the resolution of the image by the specified factor. For example, if `resolution: 2`, width and height will be halved.
```
floweisshardt: ```resolution_factor```

generic_throttle/package.xml
```diff
@@ -12,5 +12,5 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <exec_depend>rospy</exec_depend>
-
+  <exec_depend>cv_bridge</exec_depend>
```
floweisshardt, line 15: missing exec depend to ```rostopic```

floweisshardt (changes requested):

generic_throttle/src/generic_throttle/generic_throttle.py
```diff
@@ -101,6 +118,19 @@ def subscriber_callback(self, data, topic_id):
...
             return
 
+        try:
+            resolution = self.topics[topic_id]['resolution']
+        except KeyError:
+            resolution = None
+
+        if resolution is not None:
+            if data._type != 'sensor_msgs/Image':
+                rospy.logwarn('Resolution option is not available for ' +
```
fmessmer: we dont want to see this warning for every non-image message comming in...
can you do the type check with a different logic...maybe even during init?

MattiaRacca: If the user specifies the resolution_factor option for a non-image message, the throttle will throw the warning only once when the first non-image message comes through (as the resolution option for that non-image topic will be set to `None` - line 131 - and the `rospy.logwarn` statement will not be reached any longer).

generic_throttle/src/generic_throttle/generic_throttle.py
```diff
@@ -123,3 +153,32 @@ def _populate_dictionary(self):
...
+    def _resize_image(self, data, resolution):
+        old_image = None
+
+        try:
+            old_image = self.bridge.imgmsg_to_cv2(data)
+        except CvBridgeError as e:
+            print(e)
+            exit(20)
+
+        factor = 1.0 / resolution
```
fmessmer: if the parameter is now called factor already then there should not be another 1/factor...it makes it confusing...
if you want the resolution to be halfed...you should set the resolution_factor param to 0.5 in the yaml...and not to 2!

fmessmer (changes requested):

generic_throttle/src/generic_throttle/generic_throttle.py
fmessmer: also we need a division by 0 handling?

fmessmer (approved):
thanks @MattiaRacca for the contribution

I leave this to @ipa-fmw for final testing and merging

floweisshardt (approved):

PR #200, opened 2017-08-03
enhance ntp_monitor

fmessmer:
required for https://github.com/ipa320/cob_robots/pull/698

 - [x]  made ntp_monitor configurable via yaml
 - [x] ros.Timer for diagnostics publishing at 1 Hz
 - [x] ros.Timer for querying `ntpdate -q`
 - [x] enable/disable self-offset

### DIFF
--- a/cob_monitoring/src/ntp_monitor.py
+++ b/cob_monitoring/src/ntp_monitor.py
@@ -35,7 +35,6 @@ NAME = 'ntp_monitor'
 
 def ntp_monitor(ntp_hostname, offset=500, self_offset=500, diag_hostname = None, error_offset = 5000000):
     pub = rospy.Publisher("/diagnostics", DiagnosticArray,queue_size=50 )
-    rospy.init_node(NAME, anonymous=True)
 
     hostname = socket.gethostname()
     if diag_hostname is None:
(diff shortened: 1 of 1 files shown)


cob_monitoring/src/ntp_monitor.py
```diff
@@ -136,14 +137,30 @@ def ntp_monitor_main(argv=sys.argv):
...
         error_offset = int(options.error_offset_tol)
     except:
         parser.error("Offsets must be numbers")
+        print('Offsets must be numbers', file=sys.stderr)
+        sys.exit(0)
+
+    try:
+        rospy.init_node(NAME, anonymous=True)
+    except rospy.exceptions.ROSInitException:
+        print('NTP monitor is unable to initialize node. Master may not be running.', file=sys.stderr)
```
floweisshardt: is this really neccessary? I think we can assume that a roscore is already running as we do for all other nodes.

And in that case I'd like to use rospy.logerr/warn/info/debug for all print output.

floweisshardt (changes requested):

fmessmer:
@ipa-fmw ready to merge!

benmaidel:
@ipa-fxm you could consider using https://pypi.python.org/pypi/ntplib.

floweisshardt (approved):

PR #205, opened 2017-09-09
adding wlan_monitor

### LINKED ISSUE #203: [cob_monitoring] reanimate wifi monitor
@ipa-bnm do you remember the terminal command we used to get the signal quality?

 additionally to the signal quality it would be good to have the MAC of the currently used access point.

fmessmer:
fixes #203

### DIFF
--- a/cob_monitoring/src/cpu_monitor.py
+++ b/cob_monitoring/src/cpu_monitor.py
@@ -17,7 +17,6 @@
 
 from __future__ import with_statement, print_function
 
-import traceback
 import sys, os, time
 import subprocess
 import string
@@ -188,7 +187,7 @@ class CPUMonitor():
         except Exception, e:
             diag_level = DiagnosticStatus.ERROR
             diag_msgs = [ 'IPMI Exception' ]
-            diag_vals = [ KeyValue(key = 'Exception', value = traceback.format_exc()) ]
+            diag_vals = [ KeyValue(key = 'Exception', value = str(e)()) ]
 
         return diag_vals, diag_msgs, diag_level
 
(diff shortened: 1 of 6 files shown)


benmaidel:
This will not work. We need to read the stats from the router and not from the pc where the node is running.

fmessmer:
@benmaidel 
I decided to implement the node in a way that it can be used both for localhost as well as via ssh
(localhost will be helpful once we activate wlan on the new NUCs...)

I also squashed to hide the password :stuck_out_tongue_winking_eye: 


the node only sets the diagnostics level to ERROR in case of an exception, i.e. if not configured correctly
while running it only adds the info as KeyValues pairs...
this will help to have an idea about WLAN status during logfile/bagwatch debugging

fmessmer:
screenshot from 2018-07-11 09-42-56



for some reason, I was not able to connect the wlan_monitor to the router (10.4.2.1) from cob4-2 - neither with user+password nor with ssh-copy-id....both times, I get "Authentication failed"...

needs more testing with the paramiko ssh API

cob_monitoring/src/cpu_monitor.py
```diff
@@ -223,10 +222,10 @@ def check_core_temps(self):
                             diag_vals.append(KeyValue(key = 'Temp '+dev[0], value = str(temp)))
 
                             if temp >= self._core_temp_error:
-                                diag_level = max(diag_level, DiagnosticStatus.ERROR)
+                                diag_level = max(diag_level, DiagnosticStatus.OK) #do not set ERROR
                                 diag_msgs.append('CPU Hot')
                             elif temp >= self._core_temp_warn:
-                                diag_level = max(diag_level, DiagnosticStatus.WARN)
+                                diag_level = max(diag_level, DiagnosticStatus.OK) #do not set WARN
```
fmessmer, line 228: @floweisshardt FYI

fmessmer:
running on paul-unity
tested with paul-hh/-ingolstadt

PR #217, opened 2018-02-14
enable setting textsize for visualization of navigation goals

HannesBachter:
- fixes nothing
 - no new source code files added


floweisshardt (changes requested):
better solution to temporarily enlarge the textsize is to use dynamic reconfigure. This way we do not need to touch the bringup structure or kill/restart

fmessmer:
:+1: for dynamic_reconfigure

fmessmer:
tested with navgoal_teacher...works like a charm...
@HannesBachter @ipa-fmw

floweisshardt (approved):

PR #225, opened 2018-07-08
[WIP] parallel auto recover

### LINKED ISSUE #224: autorecover parallel service calls
autorecover waits for the recover service on every component. If one component is not available the others will never be recovered/initialized.
Just call the services via sss with the blocking=false param and collect the handle's. When every component is recovered, check all collected handel's for success.

@fmessmer FYI

fmessmer:
fixes #224 

~~not tested yet~~
testing on cob4-19

### DIFF
--- a/cob_helper_tools/scripts/auto_init.py
+++ b/cob_helper_tools/scripts/auto_init.py
@@ -27,7 +27,7 @@ class AutoInit():
 
   def __init__(self):
     self.components = rospy.get_param('~components', {})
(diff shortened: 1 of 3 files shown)


fmessmer:
does not work yet, but stopped due to low priority

problems is described in https://github.com/ipa320/cob_command_tools/issues/224#issuecomment-404181779

fmessmer:
not needed anymore

PR #227, opened 2018-07-20
new trajectory point time calculation

### LINKED ISSUE #226: [script_server] allow to specify vel and acc for each movement
currently `sss.move` gets vel and acc from the parameter for each movement. If you like to use different vel and acc you need to `set param`, `move` and `reset param`.

an enhancement would be to enable setting vel and acc in the sss.move command directly. This needs an API update of sss.move.

CAUTION: an API update will also influence command_gui (knoeppkes)

HannesBachter:
fixes #226 

- improves sss trajectory calculation with ramp profile.
- introduces acc parameter for calculating profile

### DIFF
--- a/cob_script_server/src/simple_script_server/simple_script_server.py
+++ b/cob_script_server/src/simple_script_server/simple_script_server.py
@@ -549,6 +549,13 @@ class simple_script_server:
 		else:
 			default_vel = rospy.get_param(param_string)
 
+		param_string = self.ns_global_prefix + "/" + component_name + "/default_acc"
+		if not rospy.has_param(param_string):
+			rospy.logwarn("parameter %s does not exist on ROS Parameter Server, using default of 0.1 [rad^2/sec].",param_string)
+			default_acc = 1.0 # rad^2/s
+		else:
+			default_acc = rospy.get_param(param_string)
+
 		for point in traj:
 			point_nr = point_nr + 1
 			point_msg = JointTrajectoryPoint()
@@ -564,7 +571,7 @@ class simple_script_server:
 
 			# use hardcoded point_time if no start_pos available
 			if start_pos != []:
-				point_time = self.calculate_point_time(component_name, start_pos, point, default_vel)
+				point_time = self.calculate_point_time(component_name, start_pos, point, default_vel, default_acc)
 			else:
 				point_time = 8*point_nr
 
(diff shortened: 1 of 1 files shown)


cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -574,10 +581,27 @@ def compose_trajectory(self, component_name, parameter_name):
 			traj_msg.points.append(point_msg)
 		return (traj_msg, 0)
 
-	def calculate_point_time(self, component_name, start_pos, end_pos, default_vel):
+	def calculate_point_time(elf, component_name, start_pos, end_pos, default_vel, default_acc):
```
fmessmer: type: `elf` -> `self`

cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -574,10 +581,27 @@ def compose_trajectory(self, component_name, parameter_name):
...
+				t2 = s2 / default_vel
+				t = 2 * t1 + t2
+			else:
+				# without constant velocity phase (only acc and dec)
+				# 1st phase: accelerate from v=0 to v=default_vel with a=default_acc in t=t1
+				# 2nd phase: missing because distance is to short (we already reached the distance with the acc and dec phase)
+				# 3rd phase: decceleration (analog to 1st phase)
+				t = math.sqrt(d_max / default_acc)
+			point_time = max(t, 0.4)	# use minimal point_time
+			print point_time
```
fmessmer: remove the `print` output

fmessmer (changes requested):
has this been tested on robot?

cob_script_server/src/simple_script_server/simple_script_server.py
fmessmer: also, if you would set a default value for `default_acc` in the function header (`default_acc=1.0`) it won't be a API-change...

cob_script_server/src/simple_script_server/simple_script_server.py
floweisshardt: done in https://github.com/ipa320/cob_command_tools/pull/227/commits/c44d4450ea11ae3ff399ec458251f25fb1553d95

cob_script_server/src/simple_script_server/simple_script_server.py
floweisshardt: `calculate_point_time` is only used internally (not in external sss API), thus I removed the default again because `vel` doesn't have a default either

floweisshardt (approved):

fmessmer (approved):
has been running on `cob4-19` for a while now

PR #228, opened 2018-08-03
change ntp offset stat

HannesBachter:
found this as a local change on paul-ingolstad
@floweisshardt @benmaidel @fmessmer is that important?


fmessmer:
not needed...already fixed and merged in #205

PR #230, opened 2018-10-05
do not auto-recover based on diagnostics

fmessmer:
ref https://github.com/mojin-robotics/cob4/issues/900

we do not want to auto-recover based on diagnostics by default as this situation might point to faulty/broken hardware

### DIFF
--- a/cob_helper_tools/scripts/auto_recover.py
+++ b/cob_helper_tools/scripts/auto_recover.py
@@ -29,6 +29,8 @@ class AutoRecover():
   def __init__(self):
     now = rospy.Time.now()
     self.em_state = 0
+    self.recover_emergency = rospy.get_parm('~recover_emergency', True)
+    self.recover_diagnostics = rospy.get_parm('~recover_diagnostics', True)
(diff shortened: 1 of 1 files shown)


floweisshardt (dismissed):

cob_helper_tools/scripts/auto_recover.py
```diff
@@ -29,6 +29,8 @@ class AutoRecover():
   def __init__(self):
     now = rospy.Time.now()
     self.em_state = 0
+    self.recover_emergency = rospy.get_parm('~recover_emergency', True)
+    self.recover_diagnostics = rospy.get_parm('~recover_diagnostics', False)
```
floweisshardt: does it make sense to use `True` as default for both

fmessmer: we would need to adjust all the `components.yaml`...for testing this was easier....and it is the hardware-gentle and failure-agnostic way to do

fmessmer:
with both params defaulting to true, without changing the config, this pr is a non-intrusive change

PR #232, opened 2018-10-11
new metrics (thermal throttling, idlejitter) for cpu monitor based on netdata

Acuda:
- fixes https://github.com/mojin-robotics/msh_backup/issues/372

### DIFF
--- a/cob_monitoring/src/cpu_monitor.py
+++ b/cob_monitoring/src/cpu_monitor.py
@@ -45,10 +45,10 @@ class CPUMonitor():
         self._mem_warn = rospy.get_param('~mem_warn', 25)
         self._mem_error = rospy.get_param('~mem_error', 1)
 
-        self._check_thermal_throttling_events = rospy.get_param('~check_thermal_throttling_events', True)
+        self._check_thermal_throttling_events = rospy.get_param('~check_thermal_throttling_events', False)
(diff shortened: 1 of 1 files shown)


cob_monitoring/src/cpu_monitor.py
```diff
@@ -42,6 +45,14 @@ def __init__(self, hostname, diag_hostname):
         self._mem_warn = rospy.get_param('~mem_warn', 25)
         self._mem_error = rospy.get_param('~mem_error', 1)
 
+        self._check_thermal_throttling_events = rospy.get_param('~check_thermal_throttling_events', True)
```
fmessmer: the other `check_*` params use `False` as default in order to not produce stale/error diagnostics for robots that do not want/have the respective tools setup (ipmi, netdata,...)...we should do the same for the new throttling and jitter diagnostics

fmessmer (changes requested):
for discussion:
I like the fact that this could even be used as diagnostics for netdata itself
however, it uses tools (netdata) which cannot be set up via `rosdep install`...would it make sense to use the `syslog` directly to be independent from netdata installation (as mentioned in https://github.com/mojin-robotics/msh_backup/issues/372#issuecomment-429031108)?

fmessmer (approved):
tested on robot...

PR #234, opened 2018-10-27
add ping_computers monitor

HannesBachter:
- fixes https://github.com/mojin-robotics/cob4/issues/939
     which is part of https://github.com/mojin-robotics/cob4/issues/875 
 - new source code files added ([APACHE license header](https://github.com/ipa320/setup/tree/master/templates) adjusted to **Mojin Robotics** was used) 

maybe the ping computers monitor could be integrated into [wlan_monitor.py](https://github.com/ipa320/cob_command_tools/blob/indigo_dev/cob_monitoring/src/wlan_monitor.py) or [wifi_monitor.py](https://github.com/ipa320/cob_command_tools/blob/indigo_dev/cob_monitoring/src/wifi_monitor.py)

### DIFF
--- a/cob_monitoring/src/ping_computers.py
+++ b/cob_monitoring/src/ping_computers.py
@@ -17,14 +17,11 @@
 import rospy
 import subprocess
 
+from std_msgs.msg import String
 from diagnostic_msgs.msg import DiagnosticStatus, KeyValue
 
 class ping_computers:
     def __init__(self):
-        self.diagnostics_publisher = rospy.Publisher('/diagnostics', DiagnosticStatus,queue_size=1)
-        self.diagnostics_frequency = rospy.get_param('~diagostics_frequency', None)
-        rospy.Timer(rospy.Duration(self.diagnostics_frequency), self.publish_diagnostics)
-
         self.ping_frequency = rospy.get_param('~ping_frequency', None)
         rospy.Timer(rospy.Duration(self.ping_frequency), self.ping_hosts)
         self.ping_number_packages = rospy.get_param('~ping_number_packages', None)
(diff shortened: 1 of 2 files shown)


HannesBachter:
added functionality to be run on a local pc, specifying various subnets  and publishing a designated topic

cob_monitoring/src/ping_computers.py
```diff
@@ -0,0 +1,130 @@
...
+        diagnostics_frequency = rospy.get_param('~diagostics_frequency', None)
+        if diagnostics_frequency < 0:
+            rospy.logerr("Ping Computers: invalid frequency: %f", diagnostics_frequency)
+            exit            
+        if self.host_pc == "ROBOT":
+            # on robot, publish diagnostics
+            self.diagnostics_publisher = rospy.Publisher('/diagnostics', DiagnosticStatus,queue_size=1)
+            rospy.Timer(rospy.Duration(1.0/diagnostics_frequency), self.publish_diagnostics)
+            # use all but last digits from local IP address
+            ip_prefix = ".".join(subprocess.check_output(['hostname', '-I']).split(".")[:-1])
```
fmessmer, line 46: if you specify full IP for the hosts, then with this, the script could detect itself whether it runs locally (on robot = same ip_prefix) or remote (robowatch = ip_prefix differs) - no need for the extra parameter `host_pc`

cob_monitoring/src/ping_computers.py
```diff
@@ -0,0 +1,130 @@
...
+            rospy.logdebug("Ping Computers: computer IP: %s, IP prefix: %s",subprocess.check_output(['hostname', '-I']), ip_prefix)
+            for host in self.hosts:
+                host_ip = ip_prefix+"."+str(host)
```
fmessmer: I think `hosts` should contain the full IP we want to ping - again, this helps combining local/remote variants without if-else-mechanism (see above)

cob_monitoring/src/ping_computers.py
```diff
@@ -0,0 +1,130 @@
...
+
+        self.host_pc = rospy.get_param('~host_pc', None)
+        self.hosts_pingable = {}
+        diagnostics_frequency = rospy.get_param('~diagostics_frequency', None)
+        if diagnostics_frequency < 0:
+            rospy.logerr("Ping Computers: invalid frequency: %f", diagnostics_frequency)
+            exit            
+        if self.host_pc == "ROBOT":
+            # on robot, publish diagnostics
+            self.diagnostics_publisher = rospy.Publisher('/diagnostics', DiagnosticStatus,queue_size=1)
```
fmessmer: the topic type of `/diagnostics` is 
```
rostopic info /diagnostics
Type: diagnostic_msgs/DiagnosticArray
```

cob_monitoring/src/ping_computers.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def ping(self, host):
+        """
+        Returns True if host (str) responds to a ping request.
+        """
+        rospy.logdebug("Ping Computers: pinging %s", host)
+        # Building the command. e.g.: "ping -c 1 google.com"
+        command = ['ping', '-c', str(self.ping_number_packages), host, '-q']
+        # if ping returns 0, ping was successful
+        return subprocess.call(command) == 0
```
fmessmer: I guess we could get more information out of this and not just true/false, e.g. 
```
ping -c 1 10.4.2.1
PING 10.4.2.1 (10.4.2.1) 56(84) bytes of data.
64 bytes from 10.4.2.1: icmp_seq=1 ttl=63 time=25.2 ms

--- 10.4.2.1 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 25.239/25.239/25.239/0.000 ms
```
if you parse the response of the `subprocess call` then we could put metrics such as package loss and delay into diagnostics

just create one DiagnosticStatus for each pc and append them to `DiagnosticArray`

cob_monitoring/src/ping_computers.py
```diff
@@ -0,0 +1,130 @@
...
+            ip_prefix = ".".join(subprocess.check_output(['hostname', '-I']).split(".")[:-1])
+            rospy.logdebug("Ping Computers: computer IP: %s, IP prefix: %s",subprocess.check_output(['hostname', '-I']), ip_prefix)
+            for host in self.hosts:
+                host_ip = ip_prefix+"."+str(host)
+                self.hosts_pingable.update({host_ip:False})
+        elif self.host_pc == "ROBOWATCH":
+            # publish directly to topic when running on robowatch pc
+            publish_topic = rospy.get_param('~publish_topic', None)
+            self.slack_publisher = rospy.Publisher(publish_topic, String, queue_size=1)
+            rospy.Timer(rospy.Duration(1.0/diagnostics_frequency), self.publish_slack)
```
fmessmer, line 56: no need for an extra timer...just call publish from `publish_diagnostics` - this also remove duplicate code in both functions

fmessmer (changes requested):
first quick review...I guess we haven't thought through the whole usecase yet

while we might want to use both variants (locally, remotely), I'm not sure this should be done from the same script

the robowatch version, might need 
 - to monitor several robots
 - to publish to several `slack` channels respectively (could be achieved by using the [`zmq_push`](https://github.com/mojin-robotics/mojin_tools/blob/kinetic-devel/slack/scripts/zmq_push.sh) mechanism)
 - to run without ROS (in order to detect PC failures without ROS running on the robot)

cob_monitoring/src/ping_computers.py
HannesBachter: I realized, that when we have the whole IP for the hosts in the .yaml file - we cannot symlink them but have to add a separate .yaml file for each robot

cob_monitoring/src/ping_computers.py
HannesBachter: in that case, we would publish the slack message every second, which would spam our channels - I don't know if we want that info being published so frequently

fmessmer: only publish on error, then stop until ok again

fmessmer: also, if we split the two variants, then "local" would publish diagnostics and "remote" would `zmq_push`

cob_monitoring/src/ping_computers.py
```diff
@@ -0,0 +1,130 @@
...
+        command = ['ping', '-c', str(self.ping_number_packages), host, '-q']
+        # if ping returns 0, ping was successful
+        return subprocess.call(command) == 0
+
+    def publish_diagnostics(self, event):
+        stat = DiagnosticStatus()
+        stat.name = "ping computers"
+        stat.hardware_id = "None"
+
+        all_pingable = True
```
fmessmer, line 129: https://docs.python.org/2/library/functions.html#all could be used for this

fmessmer:
not planned

PR #239, opened 2019-01-07
introduce max_retries for auto_init

fmessmer:
as discussed with @HannesBachter @floweisshardt 
without this, we never get notified about components unable to get initialized (due to CAN/HW errors) as during `init` service call the diagnostic turn green for a short period of time before toggling back to warning again...
with stopping to initialize after `max_retries` the diagnostics will stay yellow and we'll get notified


HannesBachter (dismissed):
tested on cob4-21

PR #246, opened 2019-05-29
Increase field size of 'number' in ScriptState

LoyVanBeek:
Noticed an overflow in this field due to too many calls without shutdown on the SimpleScriptServer
With uint32 and an action being called every second, the SimpleScriptServer can go on for 136 years which I think should be sufficient

 - fixes https://github.com/mojin-robotics/cob4/issues/1136
 - new source code files added (proper [APACHE license header](https://github.com/ipa320/setup/tree/master/templates) **must** be used)

### DIFF
--- a/cob_script_server/msg/ScriptState.msg
+++ b/cob_script_server/msg/ScriptState.msg
@@ -1,5 +1,5 @@
 Header header
-int16 number
+uint32 number
 string function_name
 string component_name
 string parameter_name


cob_script_server/msg/ScriptState.msg
```diff
@@ -1,5 +1,5 @@
 Header header
-int16 number
+uint32 number
```
fmessmer, line 2: did we not agree to "roundtrip" the increment?

fmessmer (changes requested):

cob_script_server/msg/ScriptState.msg
LoyVanBeek: We kind of did indeed :-)

However, that requires, on the Python side in SSS, to know what the maximum value is so you can actually round-trip when the value is too large. 
AFAIK, there is no way to get the max field size on a ROS msg and I don't like hardcoding a maximum value, because then we'll still have an issue when the message definition changes somehow. The exception is aonly raised when serializing and not when assigning (because it's just a Python `int`).

That logic (although very simple) will need testing and has a small risk of some sort of race condition when receiving multiple requests in parallel (unless we use a lock on the global variable, adding a tiny bit more complexity)

If you still prefer round-tripping, can be done of course.

fmessmer:
let's keep this open while working on it...

fmessmer:
consider #247 instead

PR #248, opened 2019-07-12
Feature/reduce load

LoyVanBeek:
- Reduces CPU load from cob_monitoring/net_monitor

On my box, just running the node to publish on `/diagnostics` at 1.0Hz produced a load of ~11-12% CPU. 
This PR reduces that to 1.4%

### DIFF
--- a/cob_monitoring/src/net_monitor.py
+++ b/cob_monitoring/src/net_monitor.py
@@ -46,233 +46,230 @@ import rospy
 
 import traceback
 import threading
-from threading import Timer
-import sys, os, time
-from time import sleep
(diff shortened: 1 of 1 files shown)


benmaidel (dismissed):

HannesBachter:
fixes https://github.com/mojin-robotics/cob4/issues/1184

LoyVanBeek:
I forgot to use the optimized version in one function call and I took the liberty to PEP-8 the code (in a separate commit so it's easily reverted if that's not desirable)

floweisshardt (approved):

PR #259, opened 2019-11-04
Feature/faster sss trajectories

LoyVanBeek:
- fixes https://github.com/mojin-robotics/cob4/issues/1277

TODO in this PR:

- [x] Take velocity and acceleration of each joint into account instead of 1 velocity and 1 acceleration for all joints
- [x] Get velocity limits from URDF. <strike>Not sure yet how to get relation between joint name and index in goal position array </strike>
- [x] Allow velocity scaling

### DIFF
--- a/cob_script_server/src/simple_script_server/simple_script_server.py
+++ b/cob_script_server/src/simple_script_server/simple_script_server.py
@@ -635,29 +635,54 @@ class simple_script_server:
 		return (traj_msg, 0)
 
 	def calculate_point_time(self, start_pos, end_pos, default_vel, default_acc):
+		if isinstance(default_vel, float):
+			default_vel = numpy.array([default_vel for _ in start_pos])
+
+		if isinstance(default_acc, float):
+			default_acc = numpy.array([default_acc for _ in start_pos])
+
 		try:
-			d_max = max(list(abs(numpy.array(start_pos) - numpy.array(end_pos))))
-			t1 = default_vel / default_acc
-			s1 = default_acc / 2.0 * t1**2
-			if (2 * s1 < d_max):
-				# with constant velocity phase (acc, const vel, dec)
-				# 1st phase: accelerate from v=0 to v=default_vel with a=default_acc in t=t1
-				# 2nd phase: constante velocity with v=default_vel and t=t2
-				# 3rd phase: decceleration (analog to 1st phase)
-				s2 = d_max - 2 * s1
-				t2 = s2 / default_vel
-				t = 2 * t1 + t2
-			else:
-				# without constant velocity phase (only acc and dec)
-				# 1st phase: accelerate from v=0 to v=default_vel with a=default_acc in t=t1
-				# 2nd phase: missing because distance is to short (we already reached the distance with the acc and dec phase)
-				# 3rd phase: decceleration (analog to 1st phase)
-				t = math.sqrt(d_max / default_acc)
-			point_time = max(t, 0.4)	# use minimal point_time
+			# d_max: Distance traveled in the move by each joint
+			d_max = numpy.abs(numpy.array(start_pos) - numpy.array(end_pos))
+
+			t1 = default_vel / default_acc  # Time needed for joints to accelerate to desired velocity
+			s1 = default_acc / 2.0 * t1**2  # Distance traveled during this time
+
+			t = numpy.zeros_like(d_max)
+
+			for i in range(len(start_pos)):
+				if 2 * s1[i] < d_max[i]:
+					# If we can accelerate and decelerate in less than the total distance, then:
+					# accelerate up to speed, travel at that speed for a bit and then decelerate, a three phase trajectory:
+					# with constant velocity phase (acc, const vel, dec)
+					# 1st phase: accelerate from v=0 to v=default_vel with a=default_acc in t=t1. Need s1 distance for this
+					# 2nd phase: constant velocity with v=default_vel and t=t2
+					# 3rd phase: deceleration (analog to 1st phase). Need s1 distance for this
+					#  ^
+					#  |    __2__
+					#  |   /     \
+					# v|1 /       \ 3
+					#  | /         \
+					#  o--------------->
+					#         t     d_max
+					s2 = d_max[i] - 2 * s1[i]
+					t2 = s2[i] / default_vel[i]
+					t[i] = 2 * t1[i] + t2
+				else:
+					# If we don't have enough distance to get to full speed, then we do
+					# without constant velocity phase (only acc and dec, so a two phase trajectory)
+					# 1st phase: accelerate from v=0 to v=default_vel with a=default_acc in t=t1
+					# 2nd phase: missing because distance is to short (we already reached the distance with the acc and dec phase)
+					# 3rd phase: deceleration (analog to 1st phase)
+					t[i] = numpy.sqrt(d_max[i] / default_acc[i])
+
+			# Instead of deciding per joint if we can do a three or two-phase trajectory,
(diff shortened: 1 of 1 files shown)


fmessmer (dismissed):
looks ok, needs testing on a robot though
@floweisshardt

fmessmer:
@LoyVanBeek @floweisshardt 
what about that special hack regarding https://github.com/ipa320/cob_command_tools/blob/indigo_dev/cob_script_server/src/simple_script_server/simple_script_server.py#L621-L622 where it needs to be decided whether the next trajectory point goes in the same direction for each joint (no stopping) or - at least - one joint changes direction (needs stopping)

floweisshardt:
> @LoyVanBeek @floweisshardt
> what about that special hack regarding https://github.com/ipa320/cob_command_tools/blob/indigo_dev/cob_script_server/src/simple_script_server/simple_script_server.py#L621-L622 where it needs to be decided whether the next trajectory point goes in the same direction for each joint (no stopping) or - at least - one joint changes direction (needs stopping)

I opened another issue for that: https://github.com/mojin-robotics/cob4/issues/1304

fmessmer:
this repo has been released/tagged in the meantime, please rebase

LoyVanBeek:
@floweisshardt 
For the speed scaling: what is the API we want for this? I'd say `sss.move('arm', 'home', blocking=True, speed_scale=0.5)`
This then multiplies each joint velocity with the `speed_scale`

fmessmer:
call it `speed_factor`...default should be `1.0`

I think we should finalize this pr together with #260 
@LoyVanBeek once you have implemented your features, we should create a merged feature branch and test/finalize on hardware...

LoyVanBeek:
Another API design thing:
`simple_script_server` gets a default speed from the parameter server, one speed for all joints whether it s revolute or prismatic etc. 

Getting this from URDF is simple enough, but is likely to be different. Users of the SSS who don't know of this change will get different behavior, with a potentially much faster or slower robot. How to tackle this?

Do we want to use the `default_vel` from the param server for those joints URDF has no velocity for?

LoyVanBeek:
> I think we should finalize this pr together with #260. @LoyVanBeek once you have implemented your features, we should create a merged feature branch and test/finalize on hardware...

Sound like a good idea. What's the best way to do this? Branch off `kinetic-devel`, then merge both these branches into the new branch? Can be 1 merge commit even

cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -634,29 +644,54 @@ def compose_trajectory(self, component_name, parameter_name):
...
-				t = 2 * t1 + t2
-			else:
-				# without constant velocity phase (only acc and dec)
-				# 1st phase: accelerate from v=0 to v=default_vel with a=default_acc in t=t1
-				# 2nd phase: missing because distance is to short (we already reached the distance with the acc and dec phase)
-				# 3rd phase: decceleration (analog to 1st phase)
-				t = math.sqrt(d_max / default_acc)
-			point_time = max(t, 0.4)	# use minimal point_time
+			# d_max: Distance traveled in the move by each joint
+			d_max = numpy.abs(numpy.array(start_pos) - numpy.array(end_pos))
```
fmessmer: this is an array of distances now, not a scalar distance_max anymore...`d_max` name should be adjusted

fmessmer (changes requested):
minor

fmessmer:
giving this PR a try - in gazebo simulation:

Screenshot from 2019-11-12 10-47-19

explanation - bumps from left to right:
 - bump 1 and 2 - are `cob_command_tools/indigo_dev` behavior for `sss.move("torso","back")` and `sss.move("torso","front")`
 - bump 3 and 4 - are `cob_command_tools/feature/faster_sss_trajectories` behavior for `sss.move("torso","back")` and `sss.move("torso","front")`
 - bump 5 - is `cob_command_tools/feature/faster_sss_trajectories` behavior for `sss.move("torso","back",speed_factor=0.5)`
 - bump 6 - is `cob_command_tools/feature/faster_sss_trajectories` behavior for `sss.move("torso","front",speed_factor=0.25)`

fazit:
 - feature works in principle
 - with `speed_factor=1.0` (default) `follow_joint_trajectory` action aborts with `PATH_TOLERANCE_VIOLATED` - at least for gazebo simulation 
    :point_right: note that `follow_joint_trajectory` controller is implemented in a way that it still continues to execute the motion although the action already aborted

:warning: **this is pr is a significant behavior change when used on a robot** :warning: 
i.e. all robot movements will be much faster as the `default_vel` is significantly lower then the `velocity_limits`
this might - in the least - surprise anybody not knowing about the change and - in the worst - lead to damage to robot and customer
thus, we should implement this speedup as a **opt-in** - disabled by default - to keep behavior as is for all current `sss.move` occurences in any robot/scenario and only enable this explicitly where wanted, i.e. `mbl`
this, can be done e.g. by passing another argument to toggle between using `default_vel` or `urdf_vel` - using `default_vel` as default

@LoyVanBeek

cob_script_server/src/simple_script_server/simple_script_server.py
LoyVanBeek: The same goes for all of them actually, numpy makes it look the same of course, but `default_vel`, `default_acc`, `d_max`, `t1`, `s1`, `t` as all arrays. I think the comment is clear about it and I prefer to not make the name more complicated but keep the expressions short.

LoyVanBeek:
As for the latter (optional `urdf_velocity=True`): I was think the same thing yesterday evening, so I'll implement that.

cob_script_server/src/simple_script_server/simple_script_server.py
fmessmer: I rather meant that `d_max` is not even `max` anymore...it's just an array of joint distances...the array part is ok...

fmessmer:
@LoyVanBeek please consider https://github.com/LoyVanBeek/cob_command_tools/pull/1

cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -594,19 +599,71 @@ def compose_trajectory(self, component_name, parameter_name):
...
-			default_vel = 0.1 # rad/s
-			rospy.logwarn("parameter %s does not exist on ROS Parameter Server, using default of %f [rad/sec].",param_string,default_vel)
+		if default_vel: # passed via argument
+			rospy.logdebug("using default_vel from argument")
+			if (type(default_vel) is float) or (type(default_vel) is int):
+				default_vel = numpy.array([default_vel for _ in start_pos])
+			elif (type(default_vel) is list) and (len(default_vel) == len(start_pos)) and all(((type(item) is float) or (type(item) is int)) for item in default_vel):
+				default_vel = default_vel
+			else:
+				rospy.logerr("parameter %s has wrong format (must be float/int or list of float/int) with proper dimensions.")
```
fmessmer: https://github.com/LoyVanBeek/cob_command_tools/pull/1#discussion_r345760065

should be: `argument 'default_vel' has wrong format...`

fmessmer (changes requested):

fmessmer:
@LoyVanBeek 
are you happy with this pr?
I'll try to test and merge today - if it's the case...

LoyVanBeek:
I'm OK with the current code, good to test on real HW. 
A next PR would run a PEP-8-ifyer over the code to fix the formatting.

fmessmer (approved):
tested on `cob4-21` :heavy_check_mark: 

without explicitly setting any of the new arguments `urdf_vel`, `default_vel`, `speed_factor` the behavior is as before...

PR #260, opened 2019-11-05
proper traj velocity calculation

fmessmer:
fixes https://github.com/mojin-robotics/cob4/issues/1304

**NEEDS EXTENSIVE TESTING AND VERIFICATION**
@floweisshardt @LoyVanBeek 

- [x] tested for torso movement (`sss.move("torso",["home", "left", "home", "right", "home"])`)
- [x] tested for linear arm movement

### DIFF
--- a/cob_script_server/src/simple_script_server/simple_script_server.py
+++ b/cob_script_server/src/simple_script_server/simple_script_server.py
@@ -707,7 +707,7 @@ class simple_script_server:
 					if not is_close(curr, nxt):
 						yield curr
 					else:
-						rospy.logwarn("dropping trajectory point {} due to is close: (curr {}, nxt: {})".format(i, curr, nxt))
+						rospy.logdebug("dropping trajectory point {} due to is close: (curr {}, nxt: {})".format(i, curr, nxt))
 
 		try:
 			traj = list(unique_next(traj))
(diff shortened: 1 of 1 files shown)


cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -659,6 +662,20 @@ def calculate_point_time(self, start_pos, end_pos, default_vel, default_acc):
 			point_time = 3.0	# use default point_time
 		return point_time
 
+	def calculate_point_velocities(self, prev, curr, post):
+		# set zero velocities for last trajectory point only
+		if not post:
+			return [0]*len(curr.positions)
+		else:
+			# check sign change
+			if prev and not numpy.all(numpy.equal(numpy.sign(numpy.subtract(numpy.array(curr.positions), numpy.array(prev.positions))), numpy.sign(numpy.subtract(numpy.array(post.positions), numpy.array(curr.positions))))):
```
LoyVanBeek: A bi of reformatting would make this more readable and thus understandable.

LoyVanBeek: Also, Instead of `numpy.subtract` you should be able to do just `numpy.array(post.positions) - numpy.array(curr.positions))`. According to https://docs.scipy.org/doc/numpy/reference/generated/numpy.subtract.html it should be equivalent for element-wise operations which is what we want here AFAIK

cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -608,19 +609,12 @@ def compose_trajectory(self, component_name, parameter_name):
...
-			#if point_nr == len(traj):
-			#	point_msg.velocities = [0]*len(joint_names)
-
-			# set zero velocity and accelerations for all trajectory points
-			point_msg.velocities = [0]*len(joint_names)
-			point_msg.accelerations = [0]*len(joint_names)
-
 			# use hardcoded point_time if no start_pos available
 			if start_pos != []:
 				point_time = self.calculate_point_time(start_pos, point, default_vel, default_acc)
```
LoyVanBeek: Shouldn't the velocity calculated below be used for calculating point time?

lindemeier:
## Test on cob4-21

sss.move("torso",["home", "left", "home", "right", "home"])

### Plotting joint_states:

indigo_dev:
image



feature
image

floweisshardt:
tested with cartesian arm movement: OK

floweisshardt (dismissed):

fmessmer:
I added a "is close" condition and i reset the velocity only element-wise

however, there are still some discontinuities in the velocities...see below (from gazebo simulation!)
Screenshot from 2019-11-08 17-32-34

but I think it's a general problem in the point_time computation:
when in "home" and commanding a trajectory starting with "home" as first trajectory config, the time_from_start for the first trajectory point is `0.4`...maybe we should drop that first config if already there

fmessmer:
I made some more changes and compared the various versions in gazebo simulation for `cob4-21` executing `sss.move("torso",["home","left","home","left","right"])` - always starting from `right`

1. `indigo_dev` - setting zero velocities/accelerations for every traj point
Screenshot from 2019-11-27 14-22-04

2. `indigo_dev` - not setting any velocities/acceleration (aka. luka hack)
Screenshot from 2019-11-27 14-27-05

3. `feature/calculate_traj_point_velocities` - without setting accelerations
Screenshot from 2019-11-27 14-19-28

4. `feature/calculate_traj_point_velocities` - with setting accelerations https://github.com/ipa320/cob_command_tools/pull/260/commits/9b5bfcbac83ab7b2dd4cc8329459587cc626a3d9
Screenshot from 2019-11-27 14-13-55


I think we are getting there....the "bumps" in the velocity profile are smooth/continuous due to accelerations being set to zero
I will test on HW on Friday...

@floweisshardt @LoyVanBeek @lindemeier 
any review comments until then?

fmessmer:
testing this with `mrl-a` in simulation executing `mblscenario` SM:
:point_right: velocity profiles are smooth
Screenshot from 2019-11-29 14-59-01
Screenshot from 2019-11-29 15-04-55

seems to work...will need to test the look-and-feel on real HW!

fmessmer:
testing on HW shows:
 - MoveLin via `sss` is still a bit jerky
   - the desired velocities do not start with `0.0`
   - are the velocities for each waypoint set correctly?
Screenshot from 2019-11-29 16-43-42
Screenshot from 2019-11-29 16-53-40
 - FollowJointTrajectory controller sends desired velocities which are larger than what is allowed in HW/URDF :point_right: thus triggering `PATH_CONSTRAINTS_VIOLATED` and fjt aborts
Screenshot from 2019-11-29 16-55-54
 - `default_acc` used for `point_time_calculation` is too slow :point_right: increase!

fmessmer:
BTW - velocity profile with "mrl" hack, i.e. not setting any velocities and accelerations in the trajectory
Screenshot from 2019-11-29 15-20-19

fmessmer:
tuning default_vel and default_acc and limits:
 - https://github.com/mojin-robotics/mojin_common/pull/74
 - https://github.com/mojin-robotics/mojin_robots/pull/250

fmessmer:
maybe, we'll have to allow an opt-out argument for this feature
@floweisshardt what do you think?

fmessmer:
I added an opt-out option `stop_at_waypoints` (default: `False`)
`sss.move("torso",[[0,0],[0,1],[0,2]])` will use the new velocity computation, i.e. velocity of at waypoint is `0.5`
`sss.move("torso",[[0,0],[0,1],[0,2]], stop_at_waypoints=True)` will use the "old" velocity computation, i.e. all waypoints will use velocity `0.0`

PR #267, opened 2019-12-06
[wlan monitor] print executing user for wlan monitor error message

HannesBachter:
before, user that was supposed to be logged into (`root`) was printed

tested on cob4-21


fmessmer (approved):
FR :ok:

PR #274, opened 2020-01-31
Add new package for scenario test tools

LoyVanBeek:
fixes https://github.com/mojin-robotics/cob4/issues/1400


floweisshardt (commented):
It's nice to have this package source code documented. But do we really need to commit all files in `_build`? Isn't the configuration file `scenario_test_tools/docs/conf.py` all we need to build the docu?

fmessmer (approved):
I'll merge it as is - we might go use it when extending our scenario test coverage

PR #275, opened 2020-02-07
refactor hz monitor

fmessmer:
@floweisshardt please review
fixes first bullet of https://github.com/mojin-robotics/cob4/issues/1348#issuecomment-582511435

- simplify and bullet-proof implementation (e.g. dict vs. list,...)
- guarantee 1Hz diagnostics rate
- to not report "no messages anymore" for slow topics, i.e. hz<diagnostics_rate

### DIFF
--- a/cob_monitoring/src/hz_monitor.py
+++ b/cob_monitoring/src/hz_monitor.py
@@ -32,9 +32,9 @@ class HzTest():
             # topic to test
             self.topics = rospy.get_param('~topics')
             # expected publishing rate
-            self.hz = rospy.get_param('~hz', None)
+            self.hz = rospy.get_param('~hz')
             # margin of error allowed
-            self.hzerror = rospy.get_param('~hzerror', None)
+            self.hzerror = rospy.get_param('~hzerror')
             # length of test
             self.window_size = float(rospy.get_param('~window_size', 100))
             # name for diagnostic message
--- a/cob_monitoring/src/hz_monitor.py
+++ b/cob_monitoring/src/hz_monitor.py
@@ -40,56 +40,41 @@ class HzTest():
             # name for diagnostic message
             self.diagnostics_name = rospy.get_param('~diagnostics_name',"")
             self.diagnostics_name = self.diagnostics_name.replace('/','_')
-            # sampling rate
-            self.sampling_rate = rospy.get_param('~sampling_rate', 1)
-            if self.hz < 2*self.sampling_rate:
-                rospy.logwarn("sampling_rate is probably to low: desired hz rate should at least be double the sampling_rate. (desired hz rate: %.2f, sampling_rate: %.2f)",self.hz, self.sampling_rate)
+            self.diagnostics_rate = 1.0
         except KeyError as e:
             rospy.logerr('hztest not initialized properly. Parameter [{}] not set. debug[{}] debug[{}]'.format(e, rospy.get_caller_id(), rospy.resolve_name(e.args[0])))
             sys.exit(1)
 
-        self.pub_diagnostics = rospy.Publisher('~diagnostics', DiagnosticArray, queue_size = 1)
+        self.rt_HZ_store = dict()
         self.missing_topics = copy.deepcopy(self.topics)
+        self.pub_diagnostics = rospy.Publisher('~diagnostics', DiagnosticArray, queue_size = 1)
+        self.timer_diagnostics = rospy.Timer(rospy.Duration(self.diagnostics_rate), self.publish_diagnostics)
+
 
     def run(self):
-        r = rospy.Rate(self.sampling_rate)
-
         # wait for first message
-        while not rospy.is_shutdown():
+        while len(self.missing_topics) != 0 and not rospy.is_shutdown():
             for topic in self.topics:
-                msg_class, real_topic, _ = rostopic.get_topic_class(topic, blocking=False) #pause hz until topic is published
+                msg_class, real_topic, msg_eval = rostopic.get_topic_class(topic, blocking=False) #pause hz until topic is published
                 if real_topic:
                     if real_topic in self.missing_topics:
                         self.missing_topics.remove(real_topic)
-            if len(self.missing_topics) == 0:
-                break
-            rospy.logdebug("hz monitor is waiting for type of topics {}.".format(self.missing_topics))
-            self.publish_diagnostics()
-            
+
             try:
-                r.sleep()
+                rospy.logdebug("hz monitor is waiting for type of topics {}.".format(self.missing_topics))
+                rospy.Rate(self.diagnostics_rate).sleep()
             except rospy.exceptions.ROSInterruptException:
                 pass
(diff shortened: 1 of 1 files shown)


cob_monitoring/src/hz_monitor.py
```diff
@@ -40,56 +40,41 @@ def __init__(self):
...
-            if len(self.missing_topics) == 0:
-                break
-            rospy.logdebug("hz monitor is waiting for type of topics %s."%str(self.missing_topics))
-            self.publish_diagnostics()
-            
+
             try:
-                r.sleep()
+                rospy.logdebug("hz monitor is waiting for type of topics {}.".format(self.missing_topics))
+                rospy.Rate(self.diagnostics_rate).sleep()
```
floweisshardt: does this really give us a constant cycle? This will freshly instantiate `rospy.Rate()`, so the information about when the last sleep ended is not taken into account.

Is this really the same as
```
r = rospy.Rate(self.diagnostics_rate)
...
r.lseep()
```

floweisshardt (changes requested):
with 
```
fmw@fmw-xps-13:~/git/robot_ws/src/cob_command_tools (refactor_hz_monitor)$ rosparam set /hz_monitor/topics [/test]
fmw@fmw-xps-13:~/git/robot_ws/src/cob_command_tools (refactor_hz_monitor)$ rostopic pub /test std_msgs/String "data: 'asd'" -r10
fmw@fmw-xps-13:~/git/robot_ws/src/cob_command_tools (refactor_hz_monitor)$ rosrun cob_monitoring hz_monitor.py 
```
and then stopping the publisher, I get the following error:
```
[INFO] [1581088523.227710]: subscribed to [/test]
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rospy/timer.py", line 235, in run
    self._callback(TimerEvent(last_expected, last_real, current_expected, current_real, last_duration))
  File "/home/fmw/git/robot_ws/src/cob_command_tools/cob_monitoring/src/hz_monitor.py", line 131, in publish_diagnostics
    and not (event.current_real.to_sec() < rt.msg_tn + 1/self.hz):  # condition to prevent WARN for topics with hz<diagnostics_rate
TypeError: unsupported operand type(s) for /: 'int' and 'NoneType'
```

cob_monitoring/src/hz_monitor.py
fmessmer: this sleep is just used at the beginning until `rostopic.get_topic_class` received the first message to determine the topic meta-data...then we break out of this `while` loop due to `len(self.missing_topics) != 0` condition

the constant cycle is guaranteed by using a Timer with `diagnostics_rate`

fmessmer:
> with
> 
> ```
> fmw@fmw-xps-13:~/git/robot_ws/src/cob_command_tools (refactor_hz_monitor)$ rosparam set /hz_monitor/topics [/test]
> fmw@fmw-xps-13:~/git/robot_ws/src/cob_command_tools (refactor_hz_monitor)$ rostopic pub /test std_msgs/String "data: 'asd'" -r10
> fmw@fmw-xps-13:~/git/robot_ws/src/cob_command_tools (refactor_hz_monitor)$ rosrun cob_monitoring hz_monitor.py 
> ```
> 
> and then stopping the publisher, I get the following error:
> 
> ```
> [INFO] [1581088523.227710]: subscribed to [/test]
> Exception in thread Thread-4:
> Traceback (most recent call last):
>   File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
>     self.run()
>   File "/opt/ros/kinetic/lib/python2.7/dist-packages/rospy/timer.py", line 235, in run
>     self._callback(TimerEvent(last_expected, last_real, current_expected, current_real, last_duration))
>   File "/home/fmw/git/robot_ws/src/cob_command_tools/cob_monitoring/src/hz_monitor.py", line 131, in publish_diagnostics
>     and not (event.current_real.to_sec() < rt.msg_tn + 1/self.hz):  # condition to prevent WARN for topics with hz<diagnostics_rate
> TypeError: unsupported operand type(s) for /: 'int' and 'NoneType'
> ```

you did not specify a desired `hz` rate and now `hzerror` epsilon for the topic...this defaulted to `None`...which of course cannot be used in a division...
I now made this a mandatory parameter and now the node exits with a proper message

floweisshardt (changes requested):
There is a constant diagnostics frequency of 1Hz: OK
But I still see the diagnostics information toggeling from `no message` to `publising rate too low`
```
rostopic pub /test std_msgs/String "data: 'asd'" -r0.5
rosparam set /hz_monitor/hzerror 1
rosparam set /hz_monitor/hz 10
rosrun cob_monitoring hz_monitor.py 
```

fmessmer:
> There is a constant diagnostics frequency of 1Hz: OK
> But I still see the diagnostics information toggeling from `no message` to `publising rate too low`
> 
> ```
> rostopic pub /test std_msgs/String "data: 'asd'" -r0.5
> rosparam set /hz_monitor/hzerror 1
> rosparam set /hz_monitor/hz 10
> rosrun cob_monitoring hz_monitor.py 
> ```

hmm, I don't think there is a good way to prevent this...
at some point we have to expect a message on a topic or consider it "no messages anymore"...

in your example, the topic is expected at a minimal rate of 10Hz-1Hz=9Hz
this is much faster than the 1Hz of the diagnostics callback
with your `rostopic pub` rate of 0.5Hz, there is always one callback iteration where `rostopic pub` published a message, resulting in "frequency too low" and one callback iteration where `rostopic pub` did not publish a message, i.e. there was 1sec without a message - although the minimal frequency of the topic specifies that there should have been 9 messages minimum within one sec - thus the hz_monitor considers this "no messages anymore"

this will always be the problem of topics with actual publish rate less then the diagnostics rate
however, they toggle between "WARN: freq too low" and "WARN: no messages"...which is not as bad as toggle between "OK" and "WARN"...

so, in my opinion, there is now way to fix this

however, the initial motivation for this PR was to prevent faulty diagnostics warnings with the gazebo simulation, where we allow topics to have `hz: 50` and `hzerror: 50`, i.e. we allow topics to have minimum frequency of 0Hz - i.e. we don't care about "frequency too low" because we know that gazebo cannot keep up with the real camera drivers publish rate...
while the gazebo topics are still published with a constant rate of about 1Hz, this toggled diagnostics between "OK" and "WARN: no messages"...which results in robot_status never being ok...which results in the scenario never being allowed to start...

so I think this PR is still a valid improvement and should be merged for the following benefits:
 - guarantee 1Hz diagnostics rate
 - fix faulty warnings with gazebo simulation
 - simplify implementation
 - unchanged behavior of other usages of hz_monitor together with other topics


@floweisshardt

floweisshardt (approved):
OK for me

PR #282, opened 2020-07-23
fix actionlib state handling for non initialized action clients

floweisshardt:

### DIFF
--- a/cob_script_server/src/simple_script_server/simple_script_server.py
+++ b/cob_script_server/src/simple_script_server/simple_script_server.py
@@ -1439,7 +1439,7 @@ class action_handle:
 		self.state_pub = rospy.Publisher("/script_server/state", ScriptState, queue_size=1)
 		self.AppendNode(blocking)
 		self.client = None
-		self.client_state = 9
+		self.client_state = GoalStatus.LOST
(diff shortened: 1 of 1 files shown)


cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -1500,7 +1501,7 @@ def get_state(self):
 		if self.client_mode != "": # not processing an actionlib client
 			return self.client_state
 		elif self.client == None:
-			return None
+			return GoalStatus.LOST
```
fmessmer, line 1503: in what context did this lead to wrong behavior/did you detect this/did you have to change this?

floweisshardt: in a test script we're calling `sss.get_state()` for a non initialized action client and check for a valid actionlib GoalStatus. So returning a valid actionlib status (`LOST`) is better then returning `None`.

fmessmer:
@floweisshardt rebased and made some additional changes...please re-test

floweisshardt (commented):
tested with emulator: OK

@fmessmer good to merge.

fmessmer (approved):

PR #286, opened 2020-10-17
fix noetic

fmessmer:


fmessmer:
needs conditional `exec_depend` after https://github.com/ros/rosdistro/pull/27017 gets merged

PR #291, opened 2021-01-04
fix python3

fmessmer:
some more fixes for python3-compatibility

@LoyVanBeek please review - is this safe? are there better ways, e.g. `six.moves`?

### DIFF
--- a/cob_command_gui/src/command_gui_buttons/command_gui_buttons.py
+++ b/cob_command_gui/src/command_gui_buttons/command_gui_buttons.py
@@ -113,7 +113,7 @@ class command_gui_buttons:
 
 	## Sorts a dictionary alphabetically
 	def SortDict(self,dictionary):
-		keys = sorted(dictionary.iterkeys())
+		keys = sorted(dictionary.keys())
 		k=[]
 		#print "keys = ", keys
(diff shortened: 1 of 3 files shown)


fmessmer:
`cob_console` not yet working :point_right: https://github.com/ipa320/cob_command_tools/issues/293

LoyVanBeek:
For the thing with `izip`, I think the current approach is OK, but there is indeed a `six.moves.zip` according to https://six.readthedocs.io/index.html?highlight=izip

Looks cleaner for sure.

fmessmer:
ToDo:
```
  File "/home/fxm/git/robot_ws/src/cob_command_tools/generic_throttle/src/generic_throttle/generic_throttle.py", line 40, in <dictcomp>
    self.topics = {item.keys()[0]: item.values()[0] for item in topics_list}
TypeError: 'dict_keys' object is not subscriptable
```

LoyVanBeek:
What is that bit of code supposed to do?

I see that `topics_list` is, well, a `List`, which does not have `dict_keys` or `values`. If it does have those, it is already a `dict` and there is no need to make a `dict` of it. 

Where can I find an example config to try this with?

PR #292, opened 2021-01-06
Fix python3

LoyVanBeek:
- fixes https://github.com/ipa320/cob_command_tools/pull/291#issuecomment-754732506


LoyVanBeek:
Should have been made against @fmessmer 's fork

PR #299, opened 2021-03-01
Fix/wlan monitor

HannesBachter:
- fixes https://github.com/mojin-robotics/cob4/issues/1627
 will now also be able to handle unconnected wifi modules:
image

Only downside is, that the `IEEE Norm` is not shown anymore


fmessmer:
the reduced info is due to using Modas AP instead of router
on cob4-12 we still see full information
Screenshot from 2021-03-02 09-50-15

fmessmer (approved):

PR #300, opened 2021-03-04
harmonize joint_states subscription

fmessmer:
minor
just found this heterogeneous handling while checking something different

### DIFF
--- a/cob_script_server/src/simple_script_server/simple_script_server.py
+++ b/cob_script_server/src/simple_script_server/simple_script_server.py
@@ -654,12 +654,12 @@ class simple_script_server:
 		# get current pos
 		timeout = 3.0
 		try:
-			joint_state = rospy.wait_for_message("/" + component_name + "/joint_states", JointState, timeout=timeout)  # type: JointState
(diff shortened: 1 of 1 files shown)


cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -654,12 +654,13 @@ def compose_trajectory(self, component_name, parameter_name, speed_factor=1.0, u
 		# get current pos
 		timeout = 3.0
 		try:
-			joint_state = rospy.wait_for_message("/" + component_name + "/joint_states", JointState, timeout=timeout)  # type: JointState
+			joint_state_topic = rospy.get_param(self.ns_global_prefix + "/" + component_name + "/joint_state_topic", "/" + component_name + "/joint_states")
```
floweisshardt, line 657: why is this needed?`convention is that each component has it's own private joint_states. See suggestion in https://github.com/mojin-robotics/mojin_robots/pull/334#issuecomment-790500868

benmaidel (approved):

cob_script_server/src/simple_script_server/simple_script_server.py
fmessmer: more details in https://github.com/mojin-robotics/mojin_robots/pull/334

PR #302, opened 2021-04-17
ROS_PYTHON_VERSION conditional dependency for python-six

mikaelarguedas:
Introspecting Python 2 dependencies in ROS Noetic, found that this package depends on `python-six` instead of its python3 counterpart.

This PR makes the dependency conditional on the ROS_PYTHON_VERSION.
SImilar to https://github.com/ipa320/cob_command_tools/pull/286

This is a semi-automated PR that has not been tested


fmessmer (approved):

fmessmer:
thanks @mikaelarguedas this might have slipped through...for other repos we already declare the `python-six` dependency like you proposed it in this PR :+1:

PR #306, opened 2021-11-04
fix cob_monitoring

fmessmer:
- fix string format
- add exception traceback

:point_right: use `format` for all strings
:point_right: add traceback to all exceptions


fmessmer:
works as expected

PR #314, opened 2022-03-02
[diagnostics] aquire all cpu data from netdata

### LINKED ISSUE #313: [cpu_monitor] acquire all stats/data from netdata
Since netdata already collects all the data that cpu_monitor might need, and cpu_monitor already uses some stats from netdata, we should move to getting all the data entirely from netdata.

pgehring:
fixes #313
 ref https://github.com/mojin-robotics/cob4/issues/2245

### DIFF
--- a/cob_monitoring/src/cpu_monitor.py
+++ b/cob_monitoring/src/cpu_monitor.py
@@ -216,7 +216,7 @@ class CPUMonitor():
 
         try:
             netdata_core_temp = query_netdata('sensors.coretemp_isa_0000_temperature', interval)
-            if netdata_core_temp:
+            if not netdata_core_temp:
                 diag_level = DiagnosticStatus.ERROR
                 diag_msgs = [ 'Core Temp Error' ]
                 diag_vals = [ KeyValue(key = 'Core Temp Error', value = 'Could not fetch data from netdata'),
--- a/cob_monitoring/src/cpu_monitor.py
+++ b/cob_monitoring/src/cpu_monitor.py
@@ -436,17 +436,19 @@ class CPUMonitor():
 
         load_dict = { DiagnosticStatus.OK: 'OK', DiagnosticStatus.WARN: 'High Load', DiagnosticStatus.ERROR: 'Error' }
 
-        num_cores = psutil.cpu_count()
         try:
+            netdata_info = query_netdata_info()
             netdata_system_cpu = query_netdata('system.cpu', interval)
             netdata_cpu_util = [query_netdata('cpu.cpu%d' % i for i in range(num_cores))]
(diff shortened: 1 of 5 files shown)


fmessmer (changes requested):
```
************* Module cpu_monitor
src/cob_command_tools/cob_monitoring/src/cpu_monitor.py:446: [E1120(no-value-for-parameter), CPUMonitor.check_cpu_util] No value for argument 'after' in function call
src/cob_command_tools/cob_monitoring/src/cpu_monitor.py:447: [E1120(no-value-for-parameter), CPUMonitor.check_cpu_util] No value for argument 'after' in function call
'run_pylint' returned with code '2' after 0 min 45 sec
The command ".industrial_ci/travis.sh" exited with 2.
```

cob_monitoring/src/netdata_tools/netdata_tools.py
```diff
@@ -0,0 +1,45 @@
...
+import requests
+import numpy
+
+def query_netdata_info():
+
+    try:
+        r = requests.get('http://127.0.0.1:19999/api/v1/info')
+    except requests.ConnectionError as ex:
+        rospy.logerr("NetData ConnectionError %r", ex)
+        return None
```
fmessmer: maybe we should return a tuple (data, error) - so if data is None we can write error into the diagnostics message

this should be done for all error cases in this `netdata_tools` module

cob_monitoring/src/cpu_monitor.py
```diff
@@ -261,82 +125,37 @@ def check_core_temps(self):
...
-                                 stdout = subprocess.PIPE,
-                                 stderr = subprocess.PIPE, shell = True)
-            stdout, stderr = p.communicate()
-            retcode = p.returncode
-            try:
-                stdout = stdout.decode()  #python3
-            except (UnicodeDecodeError, AttributeError):
-                pass
+            # get max freq
+            netdata_max_cpu_freq = query_netdata_info()
```
fmessmer: is this correct?
dont we need to extract the max_cpu_freq from the info - as is done for the num_cores?

fmessmer (changes requested):
I'll continue on this myself

cob_monitoring/src/cpu_monitor.py
pgehring: The variable name is a bit misleading. Should be of course also `netdata_info`

fmessmer (dismissed):
tested on my laptop
Screenshot from 2022-03-15 14-59-27

`cpu_monitor` also handles `sudo systemctl stop netdata.service` and `sudo systemctl restart netdata.service` correctly

PR #317, opened 2022-05-02
use right trigger axis as an analog signal for the run_factor

benmaidel:
To use this the joystick mode switch needs to be set to "X" instead of "D".

### DIFF
--- a/cob_teleop/ros/src/cob_teleop.cpp
+++ b/cob_teleop/ros/src/cob_teleop.cpp
@@ -55,6 +55,7 @@ public:
 
   //axis
   int axis_vx_, axis_vy_, axis_vz_, axis_roll_, axis_pitch_, axis_yaw_, axis_run_;
+  bool axis_run_updated_;
 
   //buttons
   //mode 1: Base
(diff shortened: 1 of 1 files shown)


cob_teleop/ros/src/cob_teleop.cpp
```diff
@@ -435,7 +436,21 @@ void CobTeleop::joy_cb(const sensor_msgs::Joy::ConstPtr &joy_msg)
...
+      if(joy_msg->axes[axis_run_] != 0)
+        axis_run_updated_ = true;
+      else
+        run_factor_ = 1.0;
+    }
+    if(axis_run_updated_)
+    {
+      run_factor_ = 1 + ((run_factor_param_ - 1.0) / -2.0) * (joy_msg->axes[axis_run_] - 1.0);
+    }
+  }
```
fmessmer, line 452: could this not be simplified to:
```suggestion
  if(joy_config_switch_.compare("X") == 0 && axis_run_ < joy_msg->axes.size())
  {
    if(joy_msg->axes[axis_run_] != 0)
    {
      run_factor_ = 1 + ((run_factor_param_ - 1.0) / -2.0) * (joy_msg->axes[axis_run_] - 1.0);
    }
    else
    {
      run_factor_ = 1.0;
    }
  }
```

then you don't need the `axis_run_updated_` variable....

fmessmer (changes requested):

cob_teleop/ros/src/cob_teleop.cpp
benmaidel: The joy driver works event based. That means that the values in the joy message are only updated when a button/axis on the joystick is pressed. The real range of the trigger buttons is from 1 (not pressed) to -1 (fully pressed). That means that the values for the trigger axis will stay 0 as long as no event for them was fired but cob_teleop expects a value of 1 when the trigger axis are not pressed. Thats why we need the `axis_run_updated_`

fmessmer: ok, clear now....

PR #320, opened 2022-06-22
fixing crashed thread if netdata data was malformed and make netdata module name for core temps in api calls configurable

Acuda:
there are in basic 3 problems on the netdata metrics:
* one bug in the code where the diagnostic thread crashes  on malformed data from netdata (fixed for load errors)
* sensor module name change (for temperature) on newer netdata versions (can now be configured)
  * old: `sensors.coretemp_isa_0000_temperature` new: `sensors.coretemp-isa-0000_temperature`
  * should maybe automatic changed regarding to netdata version that is also available by api call to http://127.0.0.1:19999/api/v1/info
* possible api bug (but at least a different behaivior) in newer versions (missing thread synchronization?) of netdata [**not solved** here but causes the 2 other problems, maybe as followup or as issue in [netdata repo](https://github.com/netdata/netdata)]
20220622_134422_Selection_001

### DIFF
--- a/cob_monitoring/src/cpu_monitor.py
+++ b/cob_monitoring/src/cpu_monitor.py
@@ -34,6 +34,7 @@ stat_dict = { DiagnosticStatus.OK: 'OK', DiagnosticStatus.WARN: 'Warning', Diagn
 class CPUMonitor():
     def __init__(self, hostname, diag_hostname):
         self._check_core_temps = rospy.get_param('~check_core_temps', False)
+        self._netdata_module_name_core_temps = rospy.get_param('~netdata_module_name_core_temps', 'sensors.coretemp_isa_0000_temperature')
 
         self._core_load_warn = rospy.get_param('~core_load_warn', 90)
         self._core_load_error = rospy.get_param('~core_load_error', 110)
@@ -89,7 +90,7 @@ class CPUMonitor():
         diag_level = DiagnosticStatus.OK
 
         try:
-            netdata_core_temp, error = query_netdata('sensors.coretemp_isa_0000_temperature', interval)
+            netdata_core_temp, error = query_netdata(self._netdata_module_name_core_temps, interval)
             if not netdata_core_temp:
                 diag_level = DiagnosticStatus.ERROR
                 diag_msgs = [ 'Core Temp Error' ]
@@ -189,11 +190,11 @@ class CPUMonitor():
             netdata_cpu_load, error = query_netdata('system.load', interval)
             if not netdata_cpu_load:
                 diag_level = DiagnosticStatus.ERROR
-                diag_msgs = [ 'Load Error' ]
+                diag_msg = 'Load Error'
                 diag_vals = [ KeyValue(key = 'Load Error', value = 'Could not fetch data from netdata'),
                               KeyValue(key = 'Output', value = netdata_cpu_load),
                               KeyValue(key = 'Error', value= error) ]
-                return (diag_vals, diag_msgs, diag_level)
+                return (diag_vals, diag_msg, diag_level)
 
             del netdata_cpu_load['time']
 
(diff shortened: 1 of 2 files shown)


fmessmer:
as an additional info, here are the versions for `10.4.22.11` and `10.4.22.22` respectively:
Screenshot from 2022-07-08 16-26-03
Screenshot from 2022-07-08 16-26-09

we should, make sure the patch works for both versions
we should, make sure all pcs have the same `netdata` version installed

LoyVanBeek:
Regarding the same versions: something funny is going on as well.
on `b1.cob4-22` (at `10.4.22.11`) the netdata util is installed in a different location than on `t2.cob4-22` (at `10.4.22.22`)

`/usr/sbin/netdata -P /var/run/netdata/netdata.pid -D` vs `/opt/netdata/bin/srv/netdata -P /opt/netdata/var/run/netdata/netdata.pid -D` respecively.

Found by 
```bash
# Get the PID of the process serving on port `19999` and the commandline of that process:
ps -jc $(sudo netstat -tlpn | grep 19999 | awk '{split($7, a, "/"); print a[1]}')
```

cob_monitoring/src/cpu_monitor.py
```diff
@@ -89,7 +90,7 @@ def check_core_temps(self, interval=1):
         diag_level = DiagnosticStatus.OK
 
         try:
-            netdata_core_temp, error = query_netdata('sensors.coretemp_isa_0000_temperature', interval)
+            netdata_core_temp, error = query_netdata(self._netdata_module_name_core_temps, interval)
```
fmessmer, line 93: can we not try both names and use the one that works...and only DiagnosticStatus.ERROR if BOTH fail
would be easier than using a parameter

cob_monitoring/src/cpu_monitor.py
```diff
@@ -189,11 +190,11 @@ def check_uptime(self, interval=1):
             netdata_cpu_load, error = query_netdata('system.load', interval)
```
fmessmer, line 190: why is the netdata system.uptime stuff mixed with netdata system.load in the first place anyways?
this should be kept separate!

fmessmer (changes requested):

fmessmer:
> Regarding the same versions: something funny is going on as well. on `b1.cob4-22` (at `10.4.22.11`) the netdata util is installed in a different location than on `t2.cob4-22` (at `10.4.22.22`)
> 
> `/usr/sbin/netdata -P /var/run/netdata/netdata.pid -D` vs `/opt/netdata/bin/srv/netdata -P /opt/netdata/var/run/netdata/netdata.pid -D` respecively.
> 
> Found by
> 
> ```shell
> # Get the PID of the process serving on port `19999` and the commandline of that process:
> ps -jc $(sudo netstat -tlpn | grep 19999 | awk '{split($7, a, "/"); print a[1]}')
> ```

b1 and t2 dont have the same version installed!
t2 definitely has a newer kernel

maybe on b1, netdata was installed via apt and on t2 via pip even...

but that should not make a difference...if both versions (or deb vs. pip) can offer the same info, then the cpu_monotor should be able to extract it!

cob_monitoring/src/cpu_monitor.py
LoyVanBeek: Agreed. Thought https://github.com/ipa320/cob_command_tools/blame/indigo_dev/cob_monitoring/src/cpu_monitor.py#L167 makes me wonder if there is a reason for these to be mixed. 
@pgehring, is there a reason for these to be mixed?

`system.load` was added to `check_uptime` in https://github.com/ipa320/cob_command_tools/commit/f01147699f3ccc65731c5a248977fc3f700dd89a#diff-78681ece294d0a1b6ac4749fc4af1a7ebd5ec5ddbd2d695e5799f1cf6d87b4ebR333

pgehring: This is mixed because it was gathered by the `uptime` command in bash. Since it is gathered by netdata it should be seperate - you're right.
There, I sticked too much to the old structure when restructuring it.

LoyVanBeek: OK, cool. I already split it up locally, now to test on `mrl-1`

fmessmer:
included in #321

PR #327, opened 2022-12-01
Collect only error and stale messages as errors

Deleh:
The diagnostic level was increased that warnings doesn't count as errors. This fixes that the topics like `/brake_release` get notified even if there are warnings in diagnostics.


fmessmer (commented):
this change seems "kevin"-specific and should only be merged once verified/tested on a non-kevin platform to make sure this does not brake anything...

floweisshardt:
works on cob4 and mrl-1

PR #328, opened 2023-01-09
Reconnect if SSH session is inactive

Deleh:
The `wlan_monitor` node doesn't reconnect if the SSH connection was closed on either side which requires a restart of the node to get diagnostics `OK` again. With the change, the node tries a reconnect.

### DIFF
--- a/cob_monitoring/src/wlan_monitor.py
+++ b/cob_monitoring/src/wlan_monitor.py
@@ -147,14 +147,12 @@ class IwConfigSSH(IwConfigParser):
             try:
                 self.connect()
             except Exception as e:
-                message = "IwConfigSSH update exception: %s" % e
+                message = "IwConfigSSH connect exception: %s" % e
                 self.stat.level = DiagnosticStatus.ERROR
                 self.stat.message = message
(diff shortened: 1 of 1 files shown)


cob_monitoring/src/wlan_monitor.py
```diff
@@ -149,6 +157,10 @@ def update(self):
                 self.stat.values = [ KeyValue(key = 'Exception', value = str(e)), KeyValue(key = 'Traceback', value = str(traceback.format_exc())) ]
                 rospy.logerr(message)
 
+                # Reconnect if connection is not active
+                if self.ssh.get_transport() is None or not self.ssh.get_transport().is_active():
```
fmessmer: we could use this condition to make the diagnostic message more specific, i.e.
```
message = "IwConfigSSH connection not active"
```

fmessmer (changes requested):

cob_monitoring/src/wlan_monitor.py
fmessmer: or add `ssh_connection_active: True/False` as an additional KeyValue

fmessmer (approved):

PR #329, opened 2023-01-11
Change NetData malformed data diagnostics level from ERROR to WARN

Deleh:
This changes the diagnostics level from `ERROR` to `WARN` in the following cases:
- Response from NetData is a list of `objects`
- Response from NetData is a list of `Nones`

In rare cases, the current NetData version, returns a list of `Nones` which triggers a diagnostics `ERROR`. If the robot status is checked at this point, e.g. by a statemachine, tasks could be returned as failed.
Since a malformed data error is not critical, it should be ok to only `WARN` about them.


fmessmer (commented):
lets create some kind of static counter inside the netdata_tools that increases once a query failed...then the respective monitors can query that counter and set diagnostics::ERROR based on the counter instead of the data
the counter is reset once useful data is retrieved from netdata

this way we only accept spurious malformed data and still raise an error when too many of them occur consecutively

Deleh:
I added a counter per NetData chart/label. If five consecutive responses of a label were malformed, an exception is thrown. This exception gets already catched by the `cpu_monitor` and communicated as diagnostics error.

fmessmer (changes requested):
this repo still has CI jobs for python2, which does not support the `f"text {oink}"` format
could you change back to the compatible `"text {}".format(oink)" format?

fmessmer (approved):

PR #331, opened 2023-05-02
allow max relative rotation of pi*2+0.1

benmaidel:
Relative rotational Motion was only allowed up to pi/2 with no obvious reason. Now a max of 2*pi+0.1 is allowed.

### DIFF
--- a/cob_script_server/src/simple_script_server/simple_script_server.py
+++ b/cob_script_server/src/simple_script_server/simple_script_server.py
@@ -937,7 +937,7 @@ class simple_script_server:
 			ah.set_failed(3, message)
 			return ah
 		max_rel_trans_step = 1.0 # [m]
-		max_rel_rot_step = math.pi/2 # [rad]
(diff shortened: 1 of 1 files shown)


floweisshardt (approved):

cob_script_server/src/simple_script_server/simple_script_server.py
```diff
@@ -937,7 +937,7 @@ def move_base_rel(self, component_name, parameter_name=[0,0,0], blocking=True):
 			ah.set_failed(3, message)
 			return ah
 		max_rel_trans_step = 1.0 # [m]
-		max_rel_rot_step = math.pi/2 # [rad]
+		max_rel_rot_step = math.pi*2+0.1 # [rad]
```
fmessmer, line 940: can you describe shortly what the problem was without this patch? just for documentation...

fmessmer (approved):

PR #332, opened 2023-06-27
WIP [motion_monitor] calculate direction based on nav_msgs/Path

muritane:


fmessmer (changes requested):
I'd rather have these new functionalities in a new `motion_monitor` package in `mojin_drivers` instead
it's easier to handle from a CI/dependency aspect...
if you/4am needs changes for the `emergency_stop_monitor`, then simply copy the current `emergency_stop_monitor` to the `motion_monitor` package, too and change it there...

PR #333, opened 2023-07-18
extend battery_monitor with additional rosparams

benmaidel:
Added rosparams for:
- color_charging
- color_warning
- color_error
- color_critical
- interval_duration_warning
- interval_duration_error
- interval_duration_critical

The default values for the rosparams are the same as before. So if the new parameters are not on the parameter server loaded the behavior of this node should be the same as before.

### DIFF
--- a/cob_monitoring/src/battery_monitor.py
+++ b/cob_monitoring/src/battery_monitor.py
@@ -45,6 +45,39 @@ class battery_monitor():
 
         self.enable_light = rospy.get_param("~enable_light", True)
         self.num_leds = rospy.get_param("~num_leds", 1)
+
+        if rospy.has_param("~color_charging"):
+            charging_color_param = rospy.get_param("~color_charging")
+        else:
+            # set to cyan as default
+            charging_color_param = [0.0, 1.0, 0.7, 0.4]
+        self.color_charging = ColorRGBA(*charging_color_param)
+
+        if rospy.has_param("~color_warning"):
+            color_warning_param = rospy.get_param("~color_warning")
+        else:
+            # set to cyan as default
+            color_warning_param = [1.0, 1.0, 0.0, 1.0]
+        self.color_warning = ColorRGBA(*color_warning_param)
+
+        if rospy.has_param("~color_error"):
+            color_error_param = rospy.get_param("~color_error")
+        else:
+            # set to cyan as default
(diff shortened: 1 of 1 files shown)


benmaidel:
tested on mrl-5 :heavy_check_mark:

cob_monitoring/src/battery_monitor.py
```diff
@@ -45,6 +45,39 @@ def __init__(self):
...
+        if rospy.has_param("~color_critical"):
+            color_critical_param = rospy.get_param("~color_critical")
+        else:
+            # set to cyan as default
+            color_critical_param = [1.0, 0.0, 0.0, 1.0]
+        self.color_critical = ColorRGBA(*color_critical_param)
+
+        self.interval_duration_warning = rospy.get_param("~interval_duration_warning", 30.0)
+        self.interval_duration_error = rospy.get_param("~interval_duration_error", 30.0)
+        self.interval_duration_critical = rospy.get_param("~interval_duration_critical", 30.0)
```
fmessmer: this was previously set to `5.0`

cob_monitoring/src/battery_monitor.py
```diff
@@ -45,6 +45,39 @@ def __init__(self):
...
+
+        if rospy.has_param("~color_critical"):
+            color_critical_param = rospy.get_param("~color_critical")
+        else:
+            # set to cyan as default
+            color_critical_param = [1.0, 0.0, 0.0, 1.0]
+        self.color_critical = ColorRGBA(*color_critical_param)
+
+        self.interval_duration_warning = rospy.get_param("~interval_duration_warning", 30.0)
+        self.interval_duration_error = rospy.get_param("~interval_duration_error", 30.0)
```
fmessmer: this was previously set to `15.0`

fmessmer (changes requested):

cob_monitoring/src/battery_monitor.py
benmaidel: copy past, nice catch

cob_monitoring/src/battery_monitor.py
benmaidel: copy past, nice catch

fmessmer (approved):

PR #335, opened 2023-08-16
fixed em topic based emergency_stop_monitor

benmaidel:
fixes https://github.com/4am-robotics/orga/issues/4657

- send a light command only if the state changes
- use the state field to decide if an em is active. => em could be active without button or laser being true.


benmaidel:
tested on mrl-5 :heavy_check_mark:

PR #330, opened 2023-02-14
[WIP] Remove Failed Counting from Netdata Interface

Deleh:
This is a quickfix for https://github.com/4am-robotics/orga/issues/3536 and should probably not be merged -> WIP.

The error counting on the `netdata_interface` got removed, If malformed data gets returned repeatingly, the diagnostics will stay in `WARNING`.

PR #338, opened 2024-01-12
Support New Netdata Swap Chart Name

Deleh:
The swap chart was renamed from `system.swap` to `mem.swap`. Both tables are queried now in the same manner as the CPU core temperatures are, ref
- https://github.com/ipa320/cob_command_tools/pull/321

### DIFF
--- a/cob_monitoring/src/cpu_monitor.py
+++ b/cob_monitoring/src/cpu_monitor.py
@@ -314,13 +314,34 @@ class CPUMonitor():
             diag_vals.append(KeyValue(key = 'Mem Free', value = str(free_mem)))
             diag_vals.append(KeyValue(key = 'Mem Buff/Cache', value = str(cache_mem)))
 
-            netdata_swp, error = self._netdata_interface.query_netdata('system.swap', interval)
+            # Netdata versions differ in chart names
+            netdata_swap_charts = ['mem.swap', 'system.swap']
+            error_count = 0
+            netdata_chart_err = ''
+            for chart in netdata_swap_charts:
+                try:
+                    netdata_swp, error = self._netdata_interface.query_netdata(chart, interval)
+
+                # Count individual connection errors for mutliple chart names
+                except requests.ConnectionError as err:
+                    error_count += 1
+                    netdata_chart_err += chart + ' '
+
+                    netdata_swp = None
+                    error = str(err)
+
+                if netdata_swp:
+                    break
+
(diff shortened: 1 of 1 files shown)


schmid-jn (changes requested):
image

Still some issues with NetData Diagnostic. Detected on mrl-7.

fmessmer:
would it make sense to also add a new KeyValue to the diagnostics which gives us the `netdata` version installed on the system?
most `netdata` diagnostic issues where related to a new/different `netdata` version...

Deleh:
> would it make sense to also add a new KeyValue to the diagnostics which gives us the `netdata` version installed on the system? most `netdata` diagnostic issues where related to a new/different `netdata` version...

Yes, good idea :+1:

Deleh:
> image
> 
> Still some issues with NetData Diagnostic. Detected on mrl-7.

This happens only on `mrl-7` which has indeed a different (and newer) netdata version than `mrl-8`. I'll have a look how to fix this, those `cpu.cpu*` tables seem to have gone completely.

Deleh:
This seems to be an issue with netdata itself which can't gather the relevant data. The relevant CPU data isn't even shown in the web interface.

__How it is__
image

__How it should be__
image

Let's try a reboot first.

Deleh:
> would it make sense to also add a new KeyValue to the diagnostics which gives us the `netdata` version installed on the system? most `netdata` diagnostic issues where related to a new/different `netdata` version...

I've added the netdata version as key-value pair to `CPU Info` for now.

benmaidel:
Are the changes backwards compatible with old netdata versions? If not we should not forget to update netdata on all robots (basf, afi)

Deleh:
> Are the changes backwards compatible with old netdata versions? If not we should not forget to update netdata on all robots (basf, afi)

The `swap` changes are backwards-compatible. The `version` was not included until now but is available on e.g. `mrl-2`.

fmessmer (commented):
@Deleh while you are on it
could you have a look at the old PR #330 
if it contains something useful maybe integrate it into #338 
if not, please close #330

Deleh:
> @Deleh while you are on it could you have a look at the old PR #330 if it contains something useful maybe integrate it into #338 if not, please close #330

#330 was a workaround for the `mrk` statemachine when we had serious problems with netdata. Malformed data should be treated as error, not as warning. I'll close it but keep the branch if we run into problems again.

schmid-jn:
Tested also on mrl-8: 
image
Looks god

schmid-jn (approved):
Looks good!

PR #344, opened 2024-07-31
feature/free_drive_mode

schmid-jn:
Add Button to enable free drive mode of ur-arm.

### DIFF
--- a/cob_teleop/ros/src/cob_teleop.cpp
+++ b/cob_teleop/ros/src/cob_teleop.cpp
@@ -29,6 +29,7 @@
 #include <sensor_msgs/JoyFeedbackArray.h>
 #include <std_msgs/Float64MultiArray.h>
 #include <std_msgs/ColorRGBA.h>
+#include <std_msgs/Bool.h>
 #include <std_srvs/Trigger.h>
 #include <XmlRpcValue.h>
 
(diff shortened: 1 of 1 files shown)


cob_teleop/ros/src/cob_teleop.cpp
```diff
@@ -369,6 +382,20 @@ void CobTeleop::updateBase()
   }
 }
 
+void CobTeleop::updateArm()
```
HannesBachter, line 385: why is the function not called e.g. `freedriveArm`?

cob_teleop/ros/src/cob_teleop.cpp
```diff
@@ -428,6 +455,20 @@ void CobTeleop::joy_cb(const sensor_msgs::Joy::ConstPtr &joy_msg)
     switch_mode();
   }
 
+  if (free_drive_button_ >= 0 && free_drive_button_ < (int)joy_msg->buttons.size() && joy_msg->buttons[free_drive_button_] == 1)
```
HannesBachter: why are you checking for `joy_msg->buttons[free_drive_button_] == 1` here and in the inside-if as well?

HannesBachter (changes requested):
otherwise cool feature :+1:

cob_teleop/ros/src/cob_teleop.cpp
schmid-jn: nicely spotted.

cob_teleop/ros/src/cob_teleop.cpp
schmid-jn: i thought maybe there will be more function added to the cob teleop connected to the arm. Than the main has two functions `updateBase` and `updateArm`

HannesBachter (approved):
tested on `mrl-q` :+1:

